Issue #10, opened 2023-10-27
Scalable tile retrieval tool


yellowcap:
We need a way to obtain tiles at scale from scenes.

- Retrieve pixel values for all 10m and 20m bands of Sentinel-2 data
- Retrieve data at different resolutions (10m to 320m for instance), 300m is close to Modis and Sentinel-3
- Be "batchable", a standallone script might be helpful for easy scaling across instances using AWS batch
- Track the bounding box, lat/lon centroid, and scale for each tile created

yellowcap:
v0 of this is complete. Will open specific issues when necessary or working on further improvements.

PR #11, opened 2023-10-27
Add first version of tile retriever.

yellowcap:
A first working version of a tiler that will create tiles over a scene based on location and dates. The tiles are in proper numpy format and each has metadata like bounds, centroid, and resolution.

This is a basis for discussion, but I am quite happy about the tiling overall and the approach. This should be very scalable, and can be switched to a different STAC provider very easily. 

Refs #10

### DIFF
--- a/tile_retriever.py
+++ b/tile_retriever.py
@@ -1,0 +1,233 @@
+"""
+Module to retrieve Sentinel-2 tiles.
+
+This approach uses STAC for scene discovery based on a date range and a location.
+The least cloudy scene is retrieved as a whole, and passed to a tiler. The tiler
+will create tiles of 256x256 pixels at different resolutions, track their boundaries,
+and centroids.
+"""
+import backoff
+import click
+import numpy
+import planetary_computer
+import pystac_client
+import stackstac
+from pystac_client.exceptions import APIError
+from rasterio.errors import RasterioIOError
+from rasterio.warp import transform_bounds
+from requests import ConnectionError, HTTPError
+
+BANDS = ["B02", "B03", "B04", "B05", "B06", "B07", "B08", "B8A", "B11", "B12", "SCL"]
+SIZE = 10980
+NODATA = 0
+# SCL classes.
+# 0: NO_DATA
+# 1: SATURATED_OR_DEFECTIVE
+# 2: DARK_AREA_PIXELS
+# 3: CLOUD_SHADOWS
+# 4: VEGETATION
+# 5: NOT_VEGETATED
+# 6: WATER
+# 7: UNCLASSIFIED
+# 8: CLOUD_MEDIUM_PROBABILITY
+# 9: CLOUD_HIGH_PROBABILITY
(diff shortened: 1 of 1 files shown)


tile_retriever.py
```diff
@@ -0,0 +1,233 @@
...
+    max_tries=10,
+    jitter=backoff.full_jitter,
+    raise_on_giveup=True,
+)
+def retrieve(items):
+    """
+    Download the data.
+    """
+    stack = stackstac.stack(
+        items, resolution=RESOLUTION, assets=BANDS, dtype="uint16", fill_value=NODATA
```
weiji14, line 89: Sure about setting the default nodata value to 0? I know uint16 only allows 0-65536, but 0 can also mean black.

tile_retriever.py
```diff
@@ -0,0 +1,233 @@
...
+    Data preparation pipeline.
+    """
+    print(start, end, lon, lat)
+    items = search(start, end, lon, lat)
+    stack = retrieve(items)
+    tiles, bounds, centroids, scales = tiler(stack)
+    print(f"Storing {len(tiles)} tiles")
+    # TODO: Make this an upload to S3.
+    numpy.savez_compressed(
+        f"/datadisk/clay/{stack.id.to_numpy()[0]}.npz",
```
weiji14, line 221: Can we have the option to set the folder path? And also create the folder if it doesn't already exist.

tile_retriever.py
```diff
@@ -0,0 +1,233 @@
...
+    jitter=backoff.full_jitter,
+    raise_on_giveup=True,
+)
+def retrieve(items):
+    """
+    Download the data.
+    """
+    stack = stackstac.stack(
+        items, resolution=RESOLUTION, assets=BANDS, dtype="uint16", fill_value=NODATA
+    )
```
weiji14, line 90: Discussion on optimizing reads at different resolutions: Instead of resampling to certain resolution, it will be faster (tens-of-miliseconds instead of ~1min) to read from the overviews (see https://github.com/gjoseph92/stackstac/issues/196#issuecomment-1375683468), but this will result in resolutions that are not round integer numbers.

Any thoughts on sticking to fixed resolutions like 10, 20, 60 versus overview-level defined resolutions? I know we discussed offline about aligning the 512x512 chips to what the Cloud-Optimized GeoTIFF is using internally for faster reads.

Side note: The default resampling with `stackstac` is Nearest Neighbour (see https://stackstac.readthedocs.io/en/v0.5.0/api/main/stackstac.stack.html#stackstac.stack.params.resampling), which is ok for optical images. It might be good to explicitly set the resampling algorithm in the code to be clearer.

yellowcap: I think we are going for 10m so no overviews necessary, as we stick to highest resolution available. If we decide to use 20m then yes we should try to optimize for the 20m overview if they exist.

yellowcap:
Closing in favor of https://github.com/Clay-foundation/model/pull/27

PR #24, opened 2023-11-09
Setup LightningCLI trainer script

weiji14:
Some boilerplate code for the [Lightning Trainer](https://lightning.ai/docs/pytorch/2.1.0/common/trainer.html) we'll be using later to train the neural network.

The Trainer is controlled using LightningCLI v2, and there's a placeholder BaseDataModule (data pipeline) and BaseLitModule (model architecture) in this Pull Request that can be modified later once we have decided on the data and model.

References:
- LightningCLI - https://lightning.ai/docs/pytorch/2.1.0/cli/lightning_cli.html
- LightningDataModule - https://lightning.ai/docs/pytorch/2.1.0/data/datamodule.html
- LightningModule - https://lightning.ai/docs/pytorch/2.1.0/common/lightning_module.html

Adapted from https://github.com/developmentseed/chabud2023/pull/8

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -1,0 +1,16 @@
+# Clay Foundation Model Modules
+
+This folder contains several LightningDataModule and LightningModule classes.
+
+## DataModules (data pipeline)
+
+- datamodule.py - Base data pipeline to read in Earth Observation chips
+
+## LightningModule (model architecture)
+
+- model.py - Base neural network model architecture
+
+## References
+
+- https://lightning.ai/docs/pytorch/2.1.0/data/datamodule.html
+- https://lightning.ai/docs/pytorch/2.1.0/common/lightning_module.html
--- a/src/model.py
+++ b/src/model.py
@@ -1,0 +1,65 @@
+"""
+Model architecture code.
+
+Code structure adapted from Lightning project seed at
+https://github.com/Lightning-AI/deep-learning-project-template
+"""
+import lightning as L
+import torch
+
+
+# %%
+class BaseLitModule(L.LightningModule):
+    """
+    Neural network for performing <task> on <dataset>.
+
+    Implemented using Lightning 2.1.
+    """
+
+    def __init__(self, lr: float = 0.001):
+        """
+        Define layers of the <model_name> model.
+
+        Parameters
+        ----------
+        lr : float
+            The learning rate for the Adam optimizer. Default is 0.001.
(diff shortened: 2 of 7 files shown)


src/README.md
weiji14, line 1: Ok to put these files under `src/`? Or do folks prefer another folder name?

trainer.py
```diff
@@ -0,0 +1,42 @@
...
+    Command-line inteface to run BaseLitModule with BaseDataModule.
+    """
+    cli = LightningCLI(
+        model_class=BaseLitModule,
+        datamodule_class=BaseDataModule,
+        save_config_callback=save_config_callback,
+        seed_everything_default=seed_everything_default,
+        trainer_defaults=trainer_defaults,
+        args=args,
+    )
```
weiji14, line 36: Open to discuss on whether LightningCLI is ok to use, or if we want to go with something a bit more customizable like Hydra (note: not exactly apples-to-apples comparison, see e.g. https://github.com/ashleve/lightning-hydra-template/issues/443#issuecomment-1249114069).

We could also stick with LightningCLI for now (since it's built-in to Lightning), and move on to something else later when there's a need.

weiji14 (commented):
Again, since we're trying to move fast, I'll leave this open for about a day for reviews, otherwise will merge directly :angel:

src/model.py
```diff
@@ -0,0 +1,65 @@
...
+Model architecture code.
+
+Code structure adapted from Lightning project seed at
+https://github.com/Lightning-AI/deep-learning-project-template
+"""
+import lightning as L
+import torch
+
+
+# %%
```
srmsoumya, line 11: Why do we have the artifact `%%` before starting a class?

trainer.py
srmsoumya: As you suggested, let us start with `lightning-cli` & if we feel constrained anytime during the project, we can switch to `hydra`.

srmsoumya (approved):

src/model.py
weiji14: These are used to make code blocks in Python scripts, and are supported by a few editors, see https://jupytext.readthedocs.io/en/latest/formats-scripts.html#the-percent-format.

weiji14:
Thanks @srmsoumya! I"ll merge this in and work on setting up Continuous Integration next.

PR #25, opened 2023-11-09
Setup GitHub Actions Continuous Integration tests

weiji14:
## What I am changing

- Add Continuous Integration tests for every Pull Request and push to the `main` branch

## How I did it

- Added an integration test under `src/tests/test_datapipe.py` to ensure LightningCLI from #24 works
- Add GitHub Actions workflow in `.github/workflows/test.yml` which runs on Ubuntu-22.04 and Python 3.11
  - Uses https://github.com/mamba-org/setup-micromamba to install dependencies
  - Note that the `conda-lock.yml` lockfile is used, since `micromamba` support it! However, we're installing `pytest` separately instead of adding it to the `environment.yml` file.
- Also added a .gitignore file

## How you can test it

- Install `pytest` and run `python -m pytest --verbose src/tests/` locally

## Related Issues

Based on https://github.com/developmentseed/chabud2023/pull/5

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,43 @@
+# This workflow will install Python dependencies and run tests on a single version of Python
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python
+
+name: Test
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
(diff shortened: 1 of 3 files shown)


weiji14:
Merging directly since unit tests pass, and we're trying to move fast. Will follow up with other CI stuff like linters which are less critical and can go through a more proper review cycle.

PR #27, opened 2023-11-10
Datacube

lillythomas:
This script collects overlapping Sentinel-1, Sentinel-2 and DEM for an AOI and time range, aligns to a common extent and merges the data arrays. 

To do:
- fix the `search_sentinel1_calc_max_area()` function
- choose the time to reset the merged dataarray time attribute to (right now it holds all of the respective times of its children arrays in the time dimension)
- add the tiler
- filter cloudiness on tile level
- filter out tiles with high percent of nodata for any specific channel


datacube.py
```diff
@@ -0,0 +1,382 @@
...
+    catalog = pystac_client.Client.open(STAC_API, modifier=pc.sign_inplace)
+
+    search: pystac_client.item_search.ItemSearch = catalog.search(
+        filter_lang="cql2-json",
+        filter={
+            "op": "and",
+            "args": [
+                {
+                    "op": "s_intersects",
+                    "args": [{"property": "geometry"}, geom_CENTROID.__geo_interface__],
```
srmsoumya, line 128: `CENTROID` is a point object, may be we can use that directly instead of `geom_CENTROID`

datacube.py
```diff
@@ -0,0 +1,382 @@
...
+    - BBOX (tuple): Bounding box coordinates in the format (minx, miny, maxx, maxy).
+    - resolution (int): Spatial resolution.
+    - epsg (int): EPSG code for the coordinate reference system.
+
+    Returns:
+    - tuple: A tuple containing xarray DataArrays for Sentinel-2, Sentinel-1, and DEM.
+    """
+    da_sen2: xr.DataArray = stackstac.stack(
+        items=s2_items[0],
+        epsg=26910,  # UTM Zone 10N
```
srmsoumya, line 276: Can we read the projection from the s2 item collection?

datacube.py
```diff
@@ -0,0 +1,382 @@
...
+    # To fix ValueError: unable to infer dtype on variable None
+    for key, val in da_sen1.coords.items():
+        if val.dtype == "object":
+            print("Deleting", key)
+            da_sen1 = da_sen1.drop_vars(names=key)
+
+    # Create xarray.Dataset datacube with VH and VV channels from SAR
+    da_vh: xr.DataArray = da_sen1.sel(band="vh", drop=True).rename("vh")
+    da_vv: xr.DataArray = da_sen1.sel(band="vv", drop=True).rename("vv")
+    ds_sen1: xr.Dataset = xr.merge(objects=[da_vh, da_vv], join="override")
```
srmsoumya, line 307: `da_sen1 = stackstac.mosaic(da_sen1, dim="time")` will create a composite of S1 tiles overlapping the S2 bbox. We can use that directly in the merge, and don't have to worry about pixel picking stratergy.

datacube.py
```diff
@@ -0,0 +1,382 @@
...
+        epsg=26910,  # UTM Zone 10N
+        bounds_latlon=BBOX,  # W, S, E, N
+        resolution=10,  # Spatial resolution of 10 metres
+        xy_coords="center",  # pixel centroid coords instead of topleft corner
+        dtype=np.float32,
+        fill_value=np.nan,
+    )
+
+    _, index = np.unique(da_dem['time'], return_index=True)  # Remove redundant time
+    da_dem = da_dem.isel(time=index)
```
srmsoumya, line 324: Similarly for DEM: `da_dem = stackstac.mosaic(da_dem, dim="time")`

datacube.py
```diff
@@ -0,0 +1,382 @@
...
+
+    Parameters:
+    - da_sen2 (xr.DataArray): xarray DataArray for Sentinel-2 data.
+    - da_sen1 (xr.DataArray): xarray DataArray for Sentinel-1 data.
+    - da_dem (xr.DataArray): xarray DataArray for DEM data.
+
+    Returns:
+    - xr.DataArray: Merged xarray DataArray.
+    """
+    da_merge = xr.merge([da_sen2, da_sen1, da_dem], compat='override')
```
srmsoumya, line 341: We can then probably use `concat` to merge all data arrays along the band dimensions.

srmsoumya (commented):
Looks super neat, the datacube pipeline is coming in well.

datacube.py
```diff
@@ -0,0 +1,382 @@
...
+            ],
+        },
+    )
+    s1_items = search.item_collection()
+    print(f"Found {len(s1_items)} Sentinel-1 items")
+
+    s1_gdf = gpd.GeoDataFrame.from_features(s1_items.to_dict())
+    # TODO: Find a way to get minimal number of S1 tiles to cover the entire S2 bbox
+    s1_gdf["overlap"] = s1_gdf.intersection(box(*BBOX)).area
+    s1_gdf.sort_values(by="overlap", inplace=True)
```
srmsoumya, line 184: Pick tiles from ascending or descending capture based on their overlap with the S2 bbox.
```
state = (s1_gdf[["sat:orbit_state", "overlap"]]
             .groupby(["sat:orbit_state"])
             .sum()
             .sort_values(by="overlap", ascending=False)
             .index[0])
```
`state = ascending OR descending`. 

Filter the s1_gdf by `state`: `s1_gdf = s1_gdf[s1_gdf["sat:orbit_state"] == state]`

datacube.py
lillythomas: Nice! Yes we can. Implemented.

datacube.py
lillythomas: I assigned this to the same projection used for the S2 item.

datacube.py
lillythomas: Word. I like this. Thanks.

datacube.py
lillythomas: Implemented as well.

datacube.py
weiji14: So `stackstac.mosaic` returns an `xarray.DataArray`, whereas `xarray.merge` returns an `xarray.Dataset`. The main difference is that `xarray.Dataset` allows us to clearly show named data variables (e.g. VV, VH, B1, B2, etc), whereas `xarray.DataArray` stores these in the 'band' dimension. Do we have a preference for either format?

datacube.py
weiji14, line 1: Do we want to have this file in the top-level directory, or in another folder?

datacube.py
```diff
@@ -0,0 +1,382 @@
...
+            "op": "and",
+            "args": [
+                {
+                    "op": "s_intersects",
+                    "args": [{"property": "geometry"}, CENTROID.__geo_interface__],
+                },
+                {"op": "anyinteracts", "args": [{"property": "datetime"}, week]},
+                {"op": "=", "args": [{"property": "collection"}, "sentinel-2-l2a"]},
+                {"op": "<=", "args": [{"property": "eo:cloud_cover"}, 50]},
+                {"op": "<=", "args": [{"property": "s2:nodata_pixel_percentage"}, 20]},
```
weiji14, line 125: The `eo:cloud_cover` and `s2:nodata_pixel_percentage` parameters should be made configurable perhaps?

datacube.py
```diff
@@ -0,0 +1,382 @@
...
+    """
+    da_merge = xr.merge([da_sen2, da_sen1, da_dem], compat="override")
+    print("Merged datarray: ", da_merge)
+    print(
+        "Time variables (S2, merged): ", da_sen2.time.values, da_merge.time.values
+    )  # da_sen1.time.values, da_dem.time.values
+    return da_merge
+
+
+def process(year1, year2, aoi, resolution, epsg):
```
weiji14, line 344: Not sure about needing to pass in an EPSG code here. For wide areas (in the West-to-East direction) that can span multiple UTM zones, wouldn't this force a single EPSG instead of multiple?

datacube.py
```diff
@@ -0,0 +1,382 @@
...
+    s1_items, s1_gdf = search_sentinel1(BBOX, catalog, week)
+    # s1_items, s1_gdf = search_sentinel1_calc_max_area(BBOX, catalog, week) # WIP
+
+    dem_items, dem_gdf = search_dem(BBOX, catalog, week, s2_items)
+
+    da_sen2, da_sen1, da_dem = make_dataarrays(
+        s2_items, s1_items, dem_items, BBOX, resolution, epsg
+    )
+
+    da_merge = merge_datarrays(da_sen2, da_sen1, da_dem)
```
weiji14, line 374: As a sanity check, could you plot each invidual band from this datacube and see if they look ok? I've had issues with striped outputs when running `xr.merge` before on Sentinel-1 and Copernicus DEM, and just want to double check.

datacube.py
srmsoumya: We are trying to overlap data from S1 & DEMs over S2 tiles. As S2 tiles are following the MGRS format, each tile will fall under a single UTM zone (this is my understanding, correct me if I am wrong).
S1 & DEMs might span across multiple UTM zones, but stackstac should reproject all of them together.

datacube.py
srmsoumya: Yes, but maybe we can make a separate PR for this where we run at scale for all configurable options?

datacube.py
srmsoumya: Having named data vars helps - `stackstac.mosaic` takes care of merging the pixels along the given dim (time in our case) which we don't have to handle at our end.
What if we create the data arrays & then `concat` them together as a dataset with vars attached?

datacube.py
lillythomas: It's an easy change to make so I this in [fd23e2c](https://github.com/Clay-foundation/model/commit/fd23e2cad093d6eea87e979bbb5f7d20c88b540e)

datacube.py
lillythomas: That makes sense to me. I guess I should switch this to just read the projection based on the MGRS tile's UTM zone instead of have a user argument for it.

datacube.py
lillythomas: Implemented

datacube.py
lillythomas: Reading the EPSG code directly from the Sentinel-2 item properties now.

datacube.py
lillythomas: No I think we can have this in a more appropriate subdir. I just put it here as we haven't decided on the structure yet. I'm happy to make the call and we can modify as we see fit.

datacube.py
lillythomas: Think that's already handled by `xr.merge`. It returns a:
```
Merged datarray:  <xarray.Dataset>
Dimensions: (time: 1, band: 14, x: 11365, y: 11363)
```

datacube.py
lillythomas: We still retain the named data vars I think. Merged data array using mosaic method for Sentinel 1:

```
Merged datarray:  <xarray.Dataset>
Dimensions:                                     (time: 1, band: 14, x: 11365,
                                                 y: 11363)
Coordinates: (12/67)
  * time                                        (time) datetime64[ns] 2018-06...
  * band                                        (band) <U4 'B02' 'B03' ... 'vv'
  * x                                           (x) float64 2.973e+05 ... 4.1...
  * y                                           (y) float64 4.202e+06 ... 4.0...
    id                                          (time) <U54 'S2B_MSIL2A_20180...
    s2:generation_time                          <U24 '2020-10-12T02:49:45.926Z'
    ...                                          ...
    sar:instrument_mode                         <U2 'IW'
    sar:product_type                            <U3 'GRD'
    s1:product_timeliness                       <U8 'Fast-24h'
    s1:resolution                               <U4 'high'
    description                                 (band) object nan ... 'Terrai...
    proj:shape                                  object {3600}
Data variables:
    stackstac-95bc7f7975c3ca621cbffb094f33316e  (time, band, y, x) float32 dask.array<chunksize=(1, 1, 1024, 1024), meta=np.ndarray>
    stackstac-80df347b660c0b771e7ba977f16777bc  (band, y, x) float32 dask.array<chunksize=(13, 1024, 1024), meta=np.ndarray>
    stackstac-602439b1960e897b8044c8056eb3ce96  (band, y, x) float32 dask.array<chunksize=(14, 1024, 1024), meta=np.ndarray>
Attributes:
    spec:        RasterSpec(epsg=32611, bounds=(297340, 4088320, 410990, 4201...
    crs:         epsg:32611
    transform:   | 10.00, 0.00, 297340.00|\n| 0.00,-10.00, 4201950.00|\n| 0.0...
    resolution:  10
```

lillythomas:
Ok. As of now, we are able to confidently generate a datacube w S2, S1 and DEM bands concatenated. The time dimension is that of the "best" S2 image in terms of cloud coverage. The S1 images are all of the same orbit state. Some validation results:

All bands:
Screen Shot 2023-11-14 at 2 33 47 PM

Time var:
Screen Shot 2023-11-14 at 2 33 55 PM

Some subset plots from each sensor:
Screen Shot 2023-11-14 at 2 34 32 PM
Screen Shot 2023-11-14 at 2 34 46 PM
Screen Shot 2023-11-14 at 2 34 57 PM
Screen Shot 2023-11-14 at 2 35 20 PM

@weiji14 @srmsoumya if looks good to you both, can merge and move on to the tiling part.

datacube.py
weiji14: Maybe put it in a subdir, will leave you up to you on the naming :slightly_smiling_face:

datacube.py
```diff
@@ -0,0 +1,436 @@
...
+        filter_lang="cql2-json",
+        filter={
+            "op": "and",
+            "args": [
+                {
+                    "op": "s_intersects",
+                    "args": [{"property": "geometry"}, geom_BBOX.__geo_interface__],
+                },
+                {"op": "anyinteracts", "args": [{"property": "datetime"}, week]},
+                {"op": "<=", "args": [{"property": "sat:orbit_state"}, "descending"]},
```
weiji14, line 213: What does less than or equal to descending mean??

datacube.py
```diff
@@ -0,0 +1,436 @@
...
+
+# EXAMPLE
+california_tile = gpd.read_file("ca.geojson")
+sample = california_tile.sample(1)
+aoi = sample.iloc[0].geometry
+cloud_cover_percentage = 50
+nodata_pixel_percentage = 20
+merged = process(
+    2017, 2023, aoi, 10, cloud_cover_percentage, nodata_pixel_percentage
+)  # Spatial resolution of 10 metres
```
weiji14, line 436: Wrap this block in `if __name__ == "__main__":` so that we can import functions from this datacube.py file without triggerring the datacube processing.
```suggestion
if __name__ == "__main__":
    california_tile = gpd.read_file("ca.geojson")
    sample = california_tile.sample(1)
    aoi = sample.iloc[0].geometry
    cloud_cover_percentage = 50
    nodata_pixel_percentage = 20
    merged = process(
        2017, 2023, aoi, 10, cloud_cover_percentage, nodata_pixel_percentage
    )  # Spatial resolution of 10 metres
```

weiji14 (commented):
Thanks Lilly, I think we should try to merge this by today. Just a few minor suggestions, and we can improve on stuff in follow-up PRs.

datacube.py
weiji14: Actually, I think @yellowcap is putting things under `scripts/` in #29, so we could follow that?

datacube.py
lillythomas: Thanks for catching. Old line that I thought I removed but it was in a different structure. Removed this line just now.,

datacube.py
lillythomas: Done

datacube.py
weiji14: Should be addressed in https://github.com/Clay-foundation/model/pull/27#issuecomment-1811506452

weiji14 (approved):
Linter is complaining a bit, but we can fix those later since the script will need some refining. Thanks again @lillythomas, and also @srmsoumya for reviewing!

Issue #28, opened 2023-11-13
Develop geographic sampling strategy based on Worldcover


yellowcap:
To have balanced data we probably need to oversample certain land cover types such as urban areas and wetlands. We will use a simplified version of the [Worldcover](https://esa-worldcover.org/en) dataset for this sampling effort.

Deliverables:

- [x] A script for doing sampling based on landcover classes
- [x] A list of locations / scenes to process for the initial data collection

brunosan:
As much as possible, please try to tweak this to add other sampling sources.

If we only sample to have all and roughly equal classes of land cover, we are teaching the model that that is the important variance to pay attention. This is not wrong, but I worry we will underindex the semantics that many applications need to learn about nature (besides land cover) and people. I.e. if possible sample also from:

For nature:
* [Biodiversity Intactness](https://planetarycomputer.microsoft.com/dataset/io-biodiversity)
* [Global Biodiversity Information Facility (GBIF)](https://planetarycomputer.microsoft.com/dataset/gbif) 

For people and human assets:
* [Population Density](https://sedac.ciesin.columbia.edu/data/set/gpw-v4-population-density-rev11/data-download)
* A derived dataset with number of [Building Footprints](https://planetarycomputer.microsoft.com/dataset/ms-buildings)

PR #29, opened 2023-11-13
Landcover based sampling strategy.

yellowcap:
The current sampling strategy is kept purposefully simple. We can expand to more criteria that are not only landcover, and potentially use a cluster based approach to replace the human biased perspective in the current implementation.

Refs https://github.com/Clay-foundation/model/issues/28

### DIFF
--- a/scripts/landcover.py
+++ b/scripts/landcover.py
@@ -69,7 +69,7 @@ def process(wd, worldcover, mgrs):
     """
     Run statistics and sampling.
     """
-    # compute_stats(wd, worldcover, mgrs)
+    compute_stats(wd, worldcover, mgrs)
     sample(wd)
 
 
@@ -88,9 +88,11 @@ def compute_stats(wd, worldcover, mgrs):
                 parts = shape(tile["geometry"]).geoms
                 pixels = []
                 for polygon in parts:
+                    bounds = from_bounds(*polygon.bounds, cover.transform)
                     pixels.append(
                         cover.read(
-                            1, window=from_bounds(*polygon.bounds, cover.transform)
+                            1,
+                            window=bounds,
                         ).ravel()
                     )
                 pixels = numpy.hstack(pixels)
@@ -127,7 +129,7 @@ def compute_stats(wd, worldcover, mgrs):
 
 def split_highest(data, column, size, pool=1000, seed=RANDOM_SEED):
     """
-    Split hightest values of a cloumn from a dataframe.
+    Split highest values of a column from a dataframe.
     """
     data.sort_values(column, ascending=False, inplace=True)
     return data[:pool].sample(size, random_state=seed)
@@ -194,6 +196,8 @@ def sample(wd):
             water,
         ]
     )
+
+    result = result.drop_duplicates(subset=["name"])
 
     result.to_file(Path(wd, "mgrs_sample.geojson", driver="GeoJSON"))
 
--- a/scripts/landcover.py
+++ b/scripts/landcover.py
@@ -1,0 +1,202 @@
+from pathlib import Path
+
+import click
+import fiona
+import geopandas
+import numpy
+import pandas
+import rasterio
+from fiona.crs import CRS
+from rasterio.windows import from_bounds
+from shapely.geometry import shape
+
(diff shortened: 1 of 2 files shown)


yellowcap:
The sampling scripts v0 are complete and should be fully reproducible. The current scripts selects 950 tiles based on landcover, making sure we capture diverse tiles, and a good representation of every class. The resulting tiles are visualizeed in the map below

image

Magenta are the selcted ones, the other ones are colored by number of land cover classes present in them.

yellowcap:
Attaching the sampled mgrs tiles as `geojson`, you can use this as input for the datacubes @lillythomas 
[mgrs_sample.zip](https://github.com/Clay-foundation/model/files/13365897/mgrs_sample.zip)

srmsoumya (approved):

scripts/landcover.py
```diff
@@ -0,0 +1,202 @@
...
+        driver="FlatGeobuf",
+        crs=WGS84,
+        schema=SCHEMA,
+    ) as colxn:
+        colxn.writerecords(result)
+
+
+def split_highest(data, column, size, pool=1000, seed=RANDOM_SEED):
+    """
+    Split hightest values of a cloumn from a dataframe.
```
weiji14, line 130: Small typo
```suggestion
    Split highest values of a column from a dataframe.
```

scripts/landcover.py
```diff
@@ -0,0 +1,202 @@
...
+    Sample the mgrs tiles based on landcover statistics.
+
+    Target: ~1000 tiles
+    Set very small counts to zero. Exclude high latitudes.
+    200 samples from the 2000 most diverse
+    50 samples from the 1000 highest for all other categories except water
+    100 samples from all tiles with water between 30% an 70% (making sure we
+    capture some, but exclude only purely water so we catch coasts)
+    """
+    data = geopandas.read_file(Path(wd, "mgrs_stats.fgb"))
```
weiji14, line 158: Could you share this `mgrs_stats.fgb` file please? I got a permission error trying to access `s3://esa-worldcover/v200/2021/map` for some reason, and downloading from https://worldcover2021.esa.int/downloader is taking a long time!

scripts/landcover.py
```diff
@@ -0,0 +1,202 @@
...
+            moss,
+            cropland,
+            trees,
+            shrubland,
+            grassland,
+            bare,
+            snow,
+            water,
+        ]
+    )
```
weiji14, line 196: Just to understand, this sampling function is independently getting the highest values for each category (plus some extra MGRS tiles for diversity and water areas), and then concatenating those rows together into a single dataframe?

Plotting your `mgrs_sample.geojson` file from https://github.com/Clay-foundation/model/pull/29#issuecomment-1812636118, I see a few cases where the exact MGRS tile is sampled more than once. E.g.:

MGRS tile 56VLM - sampled 3 times:

image

MGRS tile 17RMQ - sampled 2 times:

image

MGRS tile 32TLP - sampled 2 times:

image

The duplicates might be due to the independent random sampling per-category and then concatenation. Perhaps we could remove such duplicate rows before saving out the GeoJSON file?

scripts/landcover.py
yellowcap: Strange `s3 sync s3://esa-worldcover/v200/2021/map $wd/esa-worldcover-v200-2021-map --no-sign-request` works perfectly for me. But yes, its quite a lot of data (120GB) so no need for everyone to reproduce the stats file! Attaching below.

scripts/landcover.py
yellowcap: Good catch, will drop duplicates as part of the script.

yellowcap:
Sharing the stats file here @weiji14 
[mgrs_stats.zip](https://github.com/Clay-foundation/model/files/13377513/mgrs_stats.zip)

yellowcap:
Attaching the updated sampled mgrs tiles as `geojson`, this time without duplicates. Thanks @weiji14  for the catch. We can use this as input for the datacubes, cc @lillythomas 

[mgrs_sample.zip](https://github.com/Clay-foundation/model/files/13377598/mgrs_sample.zip)

PR #34, opened 2023-11-14
Add geopandas-base

weiji14:
Geographic pandas extensions!

Needed for datacube work in #27. Geopandas also has a dependency on shapely which is used in #29.


weiji14 (commented):
Just gonna quickly merge this in before #27

PR #36, opened 2023-11-15
Fix lint errors in datacube script

weiji14:
Fixing some pre-commit lint errors from #27, mostly by wrapping line-lengths of docstrings to under 88 characters.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,3 +13,6 @@ select = [
     "UP",   # pyupgrade
     "W",    # pycodestyle warnings
 ]
+
+[tool.ruff.lint.pylint]
+max-args = 6
--- a/scripts/datacube.py
+++ b/scripts/datacube.py
@@ -1,22 +1,36 @@
 """
 STAC Data Processing Script
 
-This Python script processes Sentinel-2, Sentinel-1, and DEM (Digital Elevation Model) data. It utilizes the Planetary Computer API for data retrieval and manipulation.
+This Python script processes Sentinel-2, Sentinel-1, and Copernicus DEM
+(Digital Elevation Model) data. It utilizes Microsoft's Planetary Computer API
+for data retrieval and manipulation.
 
 Constants:
 - STAC_API: Planetary Computer API endpoint
 - S2_BANDS: Bands used in Sentinel-2 data processing
 
 Functions:
-- random_date(start_year, end_year): Generate a random date within a specified range.
-- get_week(year, month, day): Get the week range for a given date.
-- get_conditions(year1, year2, cloud_cover_percentage): Get random conditions (date, year, month, day, cloud cover) within a specified year range.
-- search_sentinel2(week, aoi, cloud_cover_percentage, nodata_pixel_percentage): Search for Sentinel-2 items within a given week and area of interest.
-- search_sentinel1(BBOX, catalog, week): Search for Sentinel-1 items within a given bounding box, STAC catalog, and week.
-- search_dem(BBOX, catalog, epsg): Search for DEM items within a given bounding box.
-- make_dataarrays(s2_items, s1_items, dem_items, BBOX, resolution, epsg): Create xarray DataArrays for Sentinel-2, Sentinel-1, and DEM data.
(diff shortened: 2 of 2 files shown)


scripts/datacube.py
```diff
@@ -147,25 +174,15 @@ def search_sentinel2(week, aoi, cloud_cover_percentage, nodata_pixel_percentage)
...
-    best_clouds = (
-        s2_items_gdf[["eo:cloud_cover"]]
-        .groupby(["eo:cloud_cover"])
-        .sum()
-        .sort_values(by="eo:cloud_cover", ascending=True)
-        .index[0]
-    )
+    least_nodata_and_clouds = s2_items_gdf.sort_values(
+        by=["s2:nodata_pixel_percentage", "eo:cloud_cover"], ascending=True
+    ).index[0]
```
weiji14, line 179: @lillythomas, the linter was picking up that the `best_nodata` variable wasn't used. I've refactored this to sort based on lowest `s2:nodata_pixel_percentage` first, and then lowest `eo:cloud_cover`, so the resulting dataframe (searching between dates `2020-01-13/2020-01-19`) would look something like this:

```
                      datetime  s2:nodata_pixel_percentage  eo:cloud_cover
6  2020-01-17T18:47:09.024000Z                    0.000000       33.132727
5  2020-01-22T18:46:51.024000Z                    0.000017       40.740009
1  2020-02-11T18:45:11.024000Z                    0.000043        2.595613
0  2020-02-16T18:44:39.024000Z                    0.000056       10.043734
4  2020-01-27T18:46:29.024000Z                    0.000066       12.765600
2  2020-02-06T18:45:39.024000Z                    0.000136        2.524842
3  2020-02-01T18:46:11.024000Z                    0.000169        1.933268
```

And we would pick the first row here (datetime: `2020-01-17T18:47:09.024000Z`). Does this look ok? Or should we change the logic a bit (e.g. as long as the nodata_pixel_percentage is <20%, we just pick the lowest cloud cover, which would then be datetime: `2020-02-01T18:46:11.024000Z`)

pyproject.toml
```diff
@@ -13,3 +13,6 @@ select = [
     "UP",   # pyupgrade
     "W",    # pycodestyle warnings
 ]
+
+[tool.ruff.lint.pylint]
+max-args = 6
```
weiji14, line 18: Increasing the maximum allowed arguments to functions from the default of 5 to 6, to avoid these errors:

```
scripts/datacube.py:293:5: PLR0913 Too many arguments in function definition (6 > 5)
scripts/datacube.py:428:5: PLR0913 Too many arguments in function definition (6 > 5)
```

Alternatively, we could look at simplifying those functions to have less arguments.

scripts/datacube.py
lillythomas: Yes, I knew that `best_nodata` wasn't being used because I was mainly thinking that optimizing for least cloud cover is our best bet given that there is a pre-existing filter for nodata by way of `{"op": "<=", "args": [{"property": "s2:nodata_pixel_percentage"}, nodata_pixel_percentage]}` in the search query. I think this is a good alternative, but in some/many cases it won't optimize for cloud cover (as this example shows where the first row has a nontrivial amount of cloudy pixels - 33 %).

weiji14: Ok, so we would just pick the least cloudy image then, as long as the `nodata_pixel_percentage` is below the 20% threshold? I.e. simply remove the `best_nodata` variable and stick to having just `best_clouds`?

lillythomas: Yes I think the best call is for us to just remove the `best_nodata`  variable 👍

weiji14: Done at f65e34a9c4f72950b4f8a68a6ea2d4c580878ecd!

yellowcap (commented):
Looks good overall, if I can be extra picky, could we also use lowercase for some variables that are currently upper case, but are not constants?

Examples: https://github.com/Clay-foundation/model/blob/3608d6d7432d6a1c00fed62fa6003f6067c02ea5/scripts/datacube.py#L409, https://github.com/Clay-foundation/model/blob/3608d6d7432d6a1c00fed62fa6003f6067c02ea5/scripts/datacube.py#L239

yellowcap (approved):
Leave the remaining improvements for later. I will merge so I can work on this file.

PR #43, opened 2023-11-17
Fix datacube processor

yellowcap:
- Fixes the nodata frame boundary issue
- Improves the S1 mosaic, adding only as many scenes as necessary
- Simplifications and cosmetic edits

### DIFF
--- a/scripts/datacube.py
+++ b/scripts/datacube.py
@@ -10,31 +10,24 @@ Constants:
 - S2_BANDS: Bands used in Sentinel-2 data processing
 
 Functions:
-- random_date(start_year, end_year):
-      Generate a random date within a specified range.
-- get_week(year, month, day):
+- get_surrounding_days(reference, interval_days):
       Get the week range for a given date.
-- get_conditions(year1, year2, cloud_cover_percentage):
-      Get random conditions (date, year, month, day, cloud cover) within a
-      specified year range.
-- search_sentinel2(week, aoi, cloud_cover_percentage, nodata_pixel_percentage):
-      Search for Sentinel-2 items within a given week and area of interest.
-- search_sentinel1(BBOX, catalog, week):
+- search_sentinel2(date_range, aoi, cloud_cover_percentage, nodata_pixel_percentage):
+      Search for Sentinel-2 items within a given date range and area of interest.
+- search_sentinel1(bbox, catalog, date_range):
       Search for Sentinel-1 items within a given bounding box, STAC catalog,
-      and week.
-- search_dem(BBOX, catalog, epsg):
+      and date range.
+- search_dem(bbox, catalog):
       Search for DEM items within a given bounding box.
-- make_datasets(s2_items, s1_items, dem_items, BBOX, resolution, epsg):
+- make_datasets(s2_items, s1_items, dem_items, resolution):
       Create xarray Datasets for Sentinel-2, Sentinel-1, and DEM data.
-- merge_datasets(ds_sen2, ds_sen1, da_dem):
-      Merge xarray Datasets for Sentinel-2, Sentinel-1, and DEM.
-- process(year1, year2, aoi, resolution):
+- process(aoi, resolution, cloud_cover_percentage, nodata_pixel_percentage):
       Process Sentinel-2, Sentinel-1, and DEM data for a specified time range,
       area of interest, and resolution.
 """
 
 import random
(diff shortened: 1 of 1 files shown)


scripts/datacube.py
```diff
@@ -10,114 +10,65 @@
 - S2_BANDS: Bands used in Sentinel-2 data processing
 
 Functions:
```
yellowcap, line 12: I would actually propose to skip this header, it has gotten out of sync.

scripts/datacube.py
```diff
@@ -230,61 +174,76 @@ def search_sentinel1(BBOX, catalog, week):
...
-    s1_gdf = s1_gdf[s1_gdf["sat:orbit_state"] == state]
-    print("Filtered Sentinel-1 orbit state: ", s1_gdf["sat:orbit_state"].unique())
-    print("Number of scenes filtered by orbit state: ", len(s1_gdf))
+    # Add id as property to persist in gdf
+    for item in s1_items:
+        item.properties["id"] = item.id
+
+    # Try to find enough scenes with same orbit that fully overlap
+    # the S2 bbox.
+    s1_gdf = gpd.GeoDataFrame.from_features(s1_items)
```
yellowcap, line 193: This will fail if there are no matches. We need a way to handle cases where there are no S1 scenes. Maybe go back to search another date for S2?

scripts/datacube.py
```diff
@@ -230,61 +174,76 @@ def search_sentinel1(BBOX, catalog, week):
...
+        elif intersection.covers(geom_bbox):
+            # Stop adding scenes when the bbox is fully covered.
+            break
+        elif orbit != most_overlap_orbit:
+            # Ignore orbits that are not the same as the
+            # most overlapping one.
+            continue
+        else:
+            selected_item_ids.append(row.id)
+            intersection = intersection.intersection(row.geometry)
```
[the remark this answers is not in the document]
lillythomas: This is the solution:

```
    intersection = None
    orbit = None
    for index, row in s1_gdf.iterrows():
        orbit = row["sat:orbit_state"]
        if intersection is None and orbit == most_overlap_orbit:
            intersection = row.geometry
            selected_item_ids.append(row.id)
            intersection = intersection.intersection(row.geometry)
        elif orbit == most_overlap_orbit and intersection.covers(geom_bbox) == False:
            intersection = row.geometry
            selected_item_ids.append(row.id)
            intersection = intersection.intersection(row.geometry)
        elif orbit == most_overlap_orbit and intersection.covers(geom_bbox) == True:
            # Stop adding scenes when the bbox is fully covered.
            break
        else:
            pass
```

scripts/datacube.py
```diff
@@ -433,36 +345,37 @@ def process(
     Returns:
     - xr.Dataset: Merged xarray Dataset containing processed data.
     """
-
-    date, YEAR, MONTH, DAY, CLOUD = get_conditions(year1, year2, cloud_cover_percentage)
-    week = get_week(YEAR, MONTH, DAY)
-
-    catalog, s2_items, BBOX, epsg = search_sentinel2(
-        week, aoi, cloud_cover_percentage, nodata_pixel_percentage
+    year = random.randint(2017, 2023)
```
lillythomas, line 348: Don't think we should hard code years into the function. Those should be placed back as args

scripts/datacube.py
```diff
@@ -372,56 +315,25 @@ def make_datasets(s2_items, s1_items, dem_items, BBOX, resolution, epsg):
...
-    - da_dem (xr.DataArray): xarray DataArray for Copernicus DEM data.
-
-    Returns:
-    - xr.Dataset: Merged xarray Dataset containing valid data region.
-    """
-    # Merging the datasets as before
-    ds_merge = xr.merge([ds_sen2, ds_sen1, da_dem], compat="override")
-
-    # Define a mask where all variables have valid data (not NaN)
-    valid_mask = ds_merge.notnull().all(dim=["x", "y"])
```
lillythomas, line 406: Do we not still want to have a `valid` data mask as a safeguard @yellowcap ?

lillythomas (commented):
This PR introduces a bug and some variables that should not be hardcoded (see comments). It also removes a possibly important valid data safeguard. That said, for the sake of keeping things moving, I will merge it into the `tiler` branch I have open and fix the new issues on that branch.

PR #44, opened 2023-11-20
Ready for batch

yellowcap:
## What is changing
- Better retry strategy for Sentinel-1 tiles matching with Sentinel-2
- Storing data into S3 as compressed netCDF files
- Uint16 for Sentinel-2 data to save memory and disk space

## How it was done
- Loop through S2 scenes by cloud cover (least first) until a S1 scenes is found within 3 days
- Use tempfile to write a compressed netcdf and then upload.

Will add batch configs later if this works. Its currently in active testing on AWS Batch

## Refs
https://github.com/Clay-foundation/model/issues/10


scripts/datacube.py
```diff
@@ -394,22 +403,63 @@ def process(
...
+    date = np.datetime_as_string(merged.time.values)[:10]
+    mgrs = tile["name"]
+    version = "0.1"
+
+    merged = merged.compute()
+
+    convert_attrs_and_coords_objects_to_str(merged)
+    encoding = {key: {"zlib": True, "complevel": 9} for key in merged.variables.keys()}
+
+    with tempfile.NamedTemporaryFile() as dat:
```
weiji14, line 454: Not critical, but we can add a suffix of `.nc` to the temporary file.
```suggestion
    with tempfile.NamedTemporaryFile(suffix=".nc") as dat:
```

scripts/datacube.py
```diff
@@ -394,22 +403,63 @@ def process(
     return ds_merge
```
weiji14, line 403: Managed to save one MGRS tile, and tried dragging and dropping the NetCDF into QGIS, but it ends up being georeferenced to (0, 0). We can try to add the georeferencing later (using rioxarray) to make it easier to plot in QGIS.

lillythomas:
The script `stack_tile.py` works for calling this version of the datacube module and the tile module.  We just need to add the file writing to `tile.py` now.

scripts/datacube.py
```diff
@@ -394,22 +403,63 @@ def process(
...
         start_year,
         end_year,
         SPATIAL_RESOLUTION,
         CLOUD_COVER_PERCENTAGE,
         NODATA_PIXEL_PERCENTAGE,
     )
-    return merged
+    date = np.datetime_as_string(merged.time.values)[:10]
+    mgrs = tile["name"]
+    version = "0.1"
```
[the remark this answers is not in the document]
yellowcap: good idea

scripts/datacube.py
yellowcap: sounds good too I will change this for the tiling and then use the suffix for the tiles

scripts/datacube.py
yellowcap: I did quite some work on the xarrays so that they are proper rioxarrays. The algorithm now stores tiles that can be opened in Qgis and friends.

yellowcap:
Ok its working locally. Outputting compressed TIF tiles with 13 bands. I think that is even easier to use than netCDF. I'll try to run this in the cloud later today to see if it works there too.

yellowcap:
I will add a section for this in the README and add the configurations for the jobs etc about how to run it on Batch over the next few days in a separate PR

PR #45, opened 2023-11-21
Fix line-length, boolean comparison and import errors

weiji14:
Fix linter errors from #41:
- E501 Line too long
- E712 Comparison to `False` should be `cond is False` or `if not cond:`
- E402 Module level import not at top of file

### DIFF
--- a/scripts/stack_tile.py
+++ b/scripts/stack_tile.py
@@ -1,6 +1,4 @@
-import sys
-
-sys.path.append("./")
+# sys.path.append("./")
 from datacube import main
 from tile import tiler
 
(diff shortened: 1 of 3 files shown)


scripts/stack_tile.py
```diff
@@ -1,6 +1,3 @@
-import sys
-
-sys.path.append("./")
```
weiji14, line 3: Managed to run this script in the top-level directory (i.e. `model/`) using:

```
python scripts/stack_tile.py
```
which outputs:
```
Found 39 Sentinel-2 items
EPSG code based on Sentinel-2 item:  32643
Searching S1 in date range 2020-08-28/2020-09-03
Found 4 Sentinel-1 items
Most overlapped orbit:  descending
Found 4 DEM items
...
```
so probably don't need this sys.path.append line?

weiji14:
Gonna merge this in directly to keep the linter happy. More work on data pipeline continuing at #44.

PR #47, opened 2023-11-21
Implement MAE with support for position, time, latlon & channel embeddings

srmsoumya:
- [x] Add GeoViT from vit-pytorch that encodes position, time, latlon & channel embeddings
- [x] Add GeoMAE from vit-pytorch
- [x] Run a sample epoch to test the loss curve & training


srm_trainer.py
```diff
@@ -0,0 +1,78 @@
...
+        fast_dev_run=False,
+        accelerator="gpu",
+        devices=1,
+        # precision="16-mixed",
+        max_epochs=50,
+        # log_every_n_steps=100,
+        logger=CSVLogger("logs", name="geomae"),
+    )
+
+    trainer.fit(model, dm)
```
weiji14, line 35: Let me know once you've got the code to a stable state, and I can help with setting up the LightningCLI code to allow for different datamodules/models, following https://lightning.ai/docs/pytorch/2.1.0/cli/lightning_cli_intermediate_2.html#multiple-lightningdatamodules. That way we can have a single trainer file and do:

```
python trainer.py fit --data=GeoTIFFDataPipeModule --model=ViTLitModule  # from PR52
python trainer.py fit --data=ClayDataModule --model=GeoMAEModule  # this PR47
```

src/dataset.py
```diff
@@ -0,0 +1,68 @@
...
+        self.transform = transform
+
+    def read_chip(self, chip_path):
+        ts = chip_path.parent.name
+        year, month, day = map(int, ts.split("-"))
+
+        chip = rio.open(chip_path)
+        bounds = chip.bounds
+        centroid_x = (bounds.left + bounds.right) / 2
+        centroid_y = (bounds.top + bounds.bottom) / 2
```
weiji14, line 23: Coordinates from `chip.bounds` are UTM coordinates and not latlon. You would need to reproject first.

src/dataset.py
```diff
@@ -0,0 +1,68 @@
...
+
+        return {
+            "pixels": chip.read(),
+            "timestep": (year, month, day),
+            "latlon": (centroid_x, centroid_y),
```
weiji14, line 28: A tuple like (centroid_x, centroid_y) would be lonlat, not latlon. Longitude is X, latitude in Y.

weiji14 (commented):
Spotted some issues with how you're getting the coordinates from the GeoTIFF (there are UTM coordinates, not latlon or lonlat).

Heads up also that I've opened a small-ish PR to have the LightningDataModule return the spatiotemporal metdata (see https://github.com/Clay-foundation/model/pull/66), since we'll need to save the embeddings with some spatiotemporal metadata too.

src/dataset.py
yellowcap: I made an implementation to compute lat/lon centroids from bounds in the first tiler version, have a look
https://github.com/Clay-foundation/model/blob/9a41e3d356ac69d13ad6a7c06b20d1d21d4c9898/tile_retriever.py#L97

src/dataset.py
srmsoumya: Thanks @weiji14 swapping the values

srm_trainer.py
srmsoumya: Sure, I initially had issues with the cuda 12.x set up on my machine & thought it might be because of Lightning CLI.
Yes, I would like to move to CLI once the experiment is working fine.

srmsoumya:
@weiji14 @lillythomas I ran the `clay-small` model locally & was able to get a loss curve similar to vanilla MAE implementations from FAIR & vit-pytorch. Here is the wandb log: https://wandb.ai/devseed/CLAY-v0/runs/zj0su9df

srm_trainer.py
weiji14: Is it an error like `RuntimeError: GET was unable to find an engine to execute this computation`? I've been getting that on my AWS EC2 G5 instance, and had to set `export TORCH_CUDNN_V8_API_DISABLED=1` as a workaround from https://discuss.pytorch.org/t/segment-anything-model-unable-to-find-engine-to-execute-this-computation/179172.

weiji14:
Gonna tidy up a few things before merging in this PR. Specifically, moving code from:

- [ ] callbacks.py -> callback_wandb.py (after [#88](https://github.com/Clay-foundation/model/pull/88) maybe)- logs matplotlib figures to WandB
- [ ] srm_datamodule.py -> combine into datamodule.py. May need to recompute mean and std?
- [ ] srm_trainer.py -> combine into trainer.py

weiji14 (approved):
Ok @srmsoumya, I've cleaned up the code a little, and fixed all the lint errors. Would be nice to get some unit tests in at some point, but we can iterate on this later. Thanks again for implementing this Clay model for v0 :rocket:

PR #52, opened 2023-11-24
LightningDataModule to load GeoTIFF files

weiji14:
## What I am changing

- A [LightningDataModule](https://lightning.ai/docs/pytorch/2.1.0/data/datamodule.html) to load GeoTIFF data

## How I did it

- Using `torchdata` to construct the DataPipe
- GeoTIFF files are read using `rasterio`
- Train/validation split is 80%/20%

TODO:
- [x] Install `torchdata` dependency
- [x] Initial implementation of GeoTIFFDataPipeModule
- [x] Add extra parameters to control DataLoader (e.g. num_workers)
- [x] Add unit tests
- [ ] Refactor to load GeoTIFF data from s3 bucket instead of local drive
- [ ] etc

Notes:
- Have tried using `rioxarray` to read the GeoTIFFs, but seems a little slower than `rasterio`
- Also experimented with loading from NetCDF files using xarray's h5netcdf engine (about same speed as rioxarray loading GeoTIFF)
- Fastest seems to be loading from Zarr, but would require re-formatting of data, so leaving that to a future PR.

## How you can test it

- Download the GeoTIFF files from the s3 bucket (TODO add instructions)
- Run `python trainer.py fit --trainer.max_epochs=20 --trainer.precision=16-mixed --data.data_path=data --data.batch_size=32 --data.num_workers=8` locally

## Related Issues

References:
- https://zen3geo.readthedocs.io/en/v0.6.2/walkthrough.html

### DIFF
--- a/src/datamodule.py
+++ b/src/datamodule.py
@@ -30,7 +30,11 @@ class GeoTIFFDataPipeModule(L.LightningDataModule):
     Uses torchdata.
     """
 
-    def __init__(self, batch_size: int = 32):
+    def __init__(
+        self,
+        batch_size: int = 32,
+        num_workers: int = 8,
+    ):
         """
         Go from datacubes to 256x256 chips!
 
@@ -38,6 +42,9 @@ class GeoTIFFDataPipeModule(L.LightningDataModule):
         ----------
         batch_size : int
             Size of each mini-batch. Default is 32.
+        num_workers : int
+            How many subprocesses to use for data loading. 0 means that the
+            data will be loaded in the main process. Default is 8.
 
         Returns
         -------
@@ -46,6 +53,7 @@ class GeoTIFFDataPipeModule(L.LightningDataModule):
         """
         super().__init__()
         self.batch_size: int = batch_size
+        self.num_workers: int = num_workers
 
     def setup(self, stage: str | None = None):
         """
@@ -78,6 +86,7 @@ class GeoTIFFDataPipeModule(L.LightningDataModule):
         return torch.utils.data.DataLoader(
             dataset=self.datapipe_train,
             batch_size=None,  # handled in datapipe already
+            num_workers=self.num_workers,
         )
 
     def val_dataloader(self) -> torch.utils.data.DataLoader:
@@ -87,4 +96,5 @@ class GeoTIFFDataPipeModule(L.LightningDataModule):
         return torch.utils.data.DataLoader(
             dataset=self.datapipe_val,
             batch_size=None,  # handled in datapipe already
+            num_workers=self.num_workers,
         )
--- a/src/datamodule.py
+++ b/src/datamodule.py
@@ -32,6 +32,7 @@ class GeoTIFFDataPipeModule(L.LightningDataModule):
 
     def __init__(
         self,
+        data_path: str = "data/",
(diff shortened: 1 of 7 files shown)


src/datamodule.py
```diff
@@ -1,33 +1,33 @@
...
     """
-    Torch Dataset that returns tensors of size (13, 256, 256) with random
-    values.
+    Read a GeoTIFF file using rasterio into a numpy.ndarray, and convert it
+    to a torch.Tensor (float16 dtype).
     """
+    # GeoTIFF - Rasterio
+    with rasterio.open(fp=filepath) as dataset:
+        array: np.ndarray = dataset.read()
+        tensor: torch.Tensor = torch.as_tensor(data=array.astype(dtype="float16"))
```
yellowcap, line 21: Is float32 to float16 a save tansformation?

src/datamodule.py
```diff
@@ -1,33 +1,33 @@
...
-class RandomDataset(torch.utils.data.Dataset):
+def _array_to_torch(filepath: str) -> torch.Tensor:
     """
-    Torch Dataset that returns tensors of size (13, 256, 256) with random
-    values.
+    Read a GeoTIFF file using rasterio into a numpy.ndarray, and convert it
+    to a torch.Tensor (float16 dtype).
     """
+    # GeoTIFF - Rasterio
+    with rasterio.open(fp=filepath) as dataset:
```
yellowcap, line 19: Did you experiment with other file openers? Maybe the loader gets more stable if we use 
```
from skimage import io
im = io.imread(filepath)
```

or other tif specific loaders like

https://pypi.org/project/tifffile/

Could be worth a shot to see if that helps stabilizing the loader when compared to zarr.

src/datamodule.py
weiji14: There will be some loss of floating point precision, but we'll likely be using 16-bit precision training (see https://lightning.ai/docs/pytorch/2.1.0/common/precision_intermediate.html) to speed up the model training, so best to pre-emptively convert the data to float16 dtype here.

src/datamodule.py
weiji14: So `skimage`'s imread actually uses `tifffile` behind the scenes for reading TIFF files, see https://github.com/scikit-image/scikit-image/blob/441fe68b95a86d4ae2a351311a0c39a4232b6521/skimage/io/_io.py#L16-L68. I'll need to see if skimage/tifffile supports reading from s3 buckets directly like rasterio/GDAL. Found this thread https://github.com/cgohlke/tifffile/issues/125 which looks interesting.

weiji14:
> * [ ]  Refactor to load GeoTIFF data from s3 bucket instead of local drive

Decided to handle reading from s3 in a separate PR, because it was about 10x slower than reading from a local disk, even from the same us-east-1 region. Specifically, a mini-batch took about 0.2it/s when reading from s3, compared to ~2it/s from a local data folder. Might need to play with some I/O or networking related settings.

weiji14:
Again, merging directly in the interest of speed. Will refactor to try other drivers (following discussion at https://github.com/Clay-foundation/model/pull/52#discussion_r1404277273) later.

PR #54, opened 2023-11-24
Batch setup

yellowcap:
This is the version that we used to run our first deployment at scale.

### What changes
Main pipeline logic changes:

- Only concatenate S2, DEM, and S1 arrays for each tile during tililng. This saves lots of memory as S2 can be kept in Uint16.
- Use indexing on the arrays instead of the `.sel` function. Indexing is almost instant, and `sel` is slow for large arrays.
- Make the target bucket name an argument

Organisational changes:

- Add info about how to run the pipeline on batch.
- Add mgrs sample data as geopackage.


scripts/tile.py
```diff
@@ -84,22 +78,18 @@ def tiler(stack, date, mgrs):
...
-                    y=slice(
-                        stack.y.values[y_start],
-                        stack.y.values[y_end]
-                        + np.sign(stack.rio.transform()[0]) * EPSILON,
-                    ),
-                )
+                parts = [part[:, y_start:y_end, x_start:x_end] for part in stack]
+
+                # Only concat here to save memory, it converts S2 data to float
+                tile = xr.concat(parts, dim="band").rename("tile")
```
weiji14, line 84: @yellowcap or @lillythomas, can you double check the Sentinel-1 pixel values? I've been getting NaN loss values, and tracked it down to the Sentinel-1 VV and VH bands being all NaNs. Sentinel-2 and DEM bands are ok. Sample code:

```python
import os
import matplotlib.pyplot as plt
import rioxarray
import xarray as xr

os.environ["AWS_PROFILE"] = "xxx"

da = rioxarray.open_rasterio(
    filename="s3://clay-tiles-01/01/56HKH/2022-06-18/claytile-56HKH-2022-06-18-01-1.tif"
)
da = da.compute()

# %%
# Plot different bands
da.sel(band=1).plot.imshow()  # Blue
plt.show()
da.sel(band=10).plot.imshow()  # SWIR2
plt.show()
da.sel(band=11).plot.imshow()  # VH
plt.show()
da.sel(band=12).plot.imshow()  # VV
plt.show()
da.sel(band=13).plot.imshow()  # DEM
plt.show()
```

Blue band
SWIR2 band
VH band
VV band
DEM band

weiji14: Tried a few other tiles in other locations like MGRS 29SMD over Lisbon, and it seems like the DEM can come up with NaNs as well (in places over the ocean?). So maybe check through all those bands.

yellowcap: Good catch! Looks like the [nodata filter](https://github.com/Clay-foundation/model/blob/f48c01520099d7ef7ea916008573be3959c176a9/scripts/tile.py#L27) is missing checks on S1 and DEM. That should be easy to correct.

weiji14: Ok, look like Lilly will be fixing this in #60.

scripts/tile.py
```diff
@@ -84,22 +78,18 @@ def tiler(stack, date, mgrs):
...
 
                 if not filter_clouds_nodata(tile):
                     continue
 
+                counter += 1
+                if counter % 100 == 0:
+                    print(f"Counted {counter} tiles")
```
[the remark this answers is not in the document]
weiji14: Alternatively, if we can store the actual timestamp in the metadata of the GeoTIFF, that would be even better!

scripts/mgrs_sample.gpkg
weiji14, line 1: Should this geopackage be stored on git? Or just on the s3 bucket?

scripts/tile.py
```diff
@@ -115,12 +105,8 @@ def tiler(stack, date, mgrs):
                 with rasterio.open(name, "r+") as rst:
```
weiji14, line 105: At about L102 on the filename, @mattpaul had some suggestions at https://github.com/Clay-foundation/model/issues/35#issuecomment-1841585520 to:

1. Use underscore as the delimiter between MGRS/DATE/VERSION/COUNTER
2. Preface the VERSION with a v (e.g. `v0`)

Also mentioned at https://github.com/Clay-foundation/model/pull/44#discussion_r1399732116 that we could drop the hyphen in the date, i.e. YYYY-MM-DD becomes YYYYMMDD, but that didn't seem to have been applied in the initial batch.

scripts/tile.py
yellowcap: converted to issue for tracking https://github.com/Clay-foundation/model/issues/68

scripts/tile.py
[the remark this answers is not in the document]
yellowcap: https://github.com/Clay-foundation/model/issues/69
https://github.com/Clay-foundation/model/issues/70

scripts/mgrs_sample.gpkg
yellowcap: https://github.com/Clay-foundation/model/issues/71

scripts/tile.py
yellowcap: ok thanks will take that into consideration before next run.

yellowcap:
Converted remaining questions to issues and will merge now for the sake of moving fast.

PR #37, opened 2023-11-16
Initial Vision Transformer architecture with MAE decoder 

weiji14:
## What I am changing

- Initializing the neural network architecture, with a Vision Transformer (ViT) B/32 backbone and Masked Autoencoder (MAE) decoder

## How I did it

- ViT backbone/decoder architecture is from HuggingFace transformers
  - Loosely using the ViT B/32 model, but with 12 channels instead of 3.

Note:
- Have consider torchvision/lightly (see fb8d06b2cba9b5f7b3dcbb1ce09bf7c8cfa79b45), but changing input channels from 3 to 12 required a lot of hacking
  - Torchvision's ViT B/32 model - https://pytorch.org/vision/0.15/models/vision_transformer.html
  - MAE decoder and utils from [Lightly](https://github.com/lightly-ai/lightly) (a self-supervised learning library)
    - Adapted from https://docs.lightly.ai/self-supervised-learning/examples/mae.html

TODO:
- [x] Install `transformers` dependency
- [x] Initialize model architecture backbone/decoder layers
- [x] Setup training_step and forward pass
- [x] Add unit tests
- [x] Document model architecture in src/README.md

## How you can test it

- Run `python trainer.py fit --trainer.max_epochs=10` locally

## Related Issues

Working towards #3

References:
- He, K., Chen, X., Xie, S., Li, Y., Dollar, P., & Girshick, R. (2022). Masked Autoencoders Are Scalable Vision Learners. 2022 IEEE/CVF Conference on Computer Vision and Pattern Recognition (CVPR), 15979–15988. https://doi.org/10.1109/CVPR52688.2022.01553

PR #56, opened 2023-11-29
Generate embeddings via prediction loop

weiji14:
## What I am changing

- Generate embeddings from a pretrained ViT encoder, stored in an [npy](https://numpy.org/doc/1.26/reference/generated/numpy.lib.format.html#npy-format) file format.

## How I did it

- In the LightningModule's [`predict_step`](https://lightning.ai/docs/pytorch/2.1.0/common/lightning_module.html#predict-step), implement the logic to do the forward pass and save-to-npy step
- In the LightningDataModule, implement [`predict_dataloader`](https://lightning.ai/docs/pytorch/2.1.0/data/datamodule.html#predict-dataloader)
- Raw embeddings are of shape (1, 65, 768), and we take the mean of the patch embeddings (1, 64, 768) which becomes a (1, 768) shape embedding (see https://github.com/Clay-foundation/model/pull/56#discussion_r1410187302)

Illustration of how the raw embeddings are turned into the final embeddings of shape (1, 768).

Excalidraw link: https://excalidraw.com/#json=IDteKVYDAHd05wT-rCR6K,vmbiRIb5ucGiXP6R1idu4w

TODO in this PR:
- [x] Implement `predict_dataloader` and `predict_step`
- [x] Add a unit test
- [x] Refactor predict_dataloader to not rely on the validation datapipe
- [x] Compute mean of the raw patch embeddings

TODO in the future:
- [ ] Properly document how to generate/save embeddings and load them (see instructions below)
- [ ] Put some sort of spatiotemporal metadata in the filename of the output embedding
- [ ] Refactor to save only encoder weights to checkpoint instead of entire encoder/decoder network

## How you can test it

- Locally, download some GeoTIFF data into the `data/` folder, and then run:
```python
python trainer.py fit --trainer.max_epochs=10 --trainer.precision=bf16-mixed --data.data_path=data/56HKH --data.num_workers=4  # train the model
python trainer.py predict --ckpt_path=checkpoints/last.ckpt --data.batch_size=1 --trainer.limit_predict_batches=1  # generate embeddings
```
  - This should produce an `embedding_0.npy` file under the `data/embeddings/` folder. Sample files (need to unzip):
    - Final embedding of shape (1, 768): [embedding_0.npy.zip](https://github.com/Clay-foundation/model/files/13517583/embedding_0.npy.zip)
    - (for reference) Raw embedding of shape (1, 65, 768):  [embedding_0.npy.zip](https://github.com/Clay-foundation/model/files/13507373/embedding_0.npy.zip)

- Extra configuration options can be found using `python trainer.py predict --help`

To load the embeddings from the npy file:

```python
import numpy as np

array: np.ndarray = np.load(file="embedding_0.npy")
assert array.shape == (1, 768)
```

## Related Issues

Towards #3

### DIFF
--- a/src/model_vit.py
+++ b/src/model_vit.py
@@ -4,7 +4,10 @@ Model architecture code.
 Code structure adapted from Lightning project seed at
 https://github.com/Lightning-AI/deep-learning-project-template
 """
+import os
+
 import lightning as L
+import numpy as np
 import torch
 import transformers
 
@@ -125,6 +128,31 @@ class ViTLitModule(L.LightningModule):
         Logic for the neural network's validation loop.
         """
         pass
+
+    def predict_step(self, batch: torch.Tensor, batch_idx: int) -> torch.Tensor:
+        """
+        Logic for the neural network's prediction loop.
+        """
+        x: torch.Tensor = batch
+        # x: torch.Tensor = torch.randn(32, 13, 256, 256)  # BCHW
+
+        # Forward encoder
+        outputs_encoder: dict = self(x)
+
+        # Get embeddings generated from encoder
+        embeddings: torch.Tensor = outputs_encoder.last_hidden_state
+        assert embeddings.shape == torch.Size(
+            [self.B, 17, 768]  # (batch_size, sequence_length, hidden_size)
+        )
+
+        # Save embeddings in npy format
+        outfolder: str = f"{self.trainer.default_root_dir}/data/embeddings"
+        os.makedirs(name=outfolder, exist_ok=True)
+        outfile = f"{outfolder}/embedding_{batch_idx}.npy"
+        np.save(file=outfile, arr=embeddings.cpu())
+        print(f"Saved embeddings of shape {tuple(embeddings.shape)} to {outfile}")
+
+        return embeddings
 
     def configure_optimizers(self) -> torch.optim.Optimizer:
         """
--- a/src/model_vit.py
+++ b/src/model_vit.py
@@ -141,20 +141,27 @@ class ViTLitModule(L.LightningModule):
         outputs_encoder: dict = self(x)
 
         # Get embeddings generated from encoder
-        embeddings: torch.Tensor = outputs_encoder.last_hidden_state
-        assert embeddings.shape == torch.Size(
+        embeddings_raw: torch.Tensor = outputs_encoder.last_hidden_state
+        assert embeddings_raw.shape == torch.Size(
             [self.B, 65, 768]  # (batch_size, sequence_length, hidden_size)
         )
-        assert not torch.isnan(embeddings).any()  # ensure no NaNs in embedding
+        assert not torch.isnan(embeddings_raw).any()  # ensure no NaNs in embedding
+
+        # Take the mean of the embeddings along the sequence_length dimension
+        # excluding the first cls token embedding, compute over patch embeddings
+        embeddings_mean: torch.Tensor = embeddings_raw[:, 1:, :].mean(dim=1)
+        assert embeddings_mean.shape == torch.Size(
+            [self.B, 768]  # (batch_size, hidden_size)
+        )
 
         # Save embeddings in npy format
         outfolder: str = f"{self.trainer.default_root_dir}/data/embeddings"
         os.makedirs(name=outfolder, exist_ok=True)
         outfile = f"{outfolder}/embedding_{batch_idx}.npy"
-        np.save(file=outfile, arr=embeddings.cpu())
-        print(f"Saved embeddings of shape {tuple(embeddings.shape)} to {outfile}")
+        np.save(file=outfile, arr=embeddings_mean.cpu())
+        print(f"Saved embeddings of shape {tuple(embeddings_mean.shape)} to {outfile}")
 
-        return embeddings
+        return embeddings_mean
 
     def configure_optimizers(self) -> torch.optim.Optimizer:
         """
--- a/src/model_vit.py
+++ b/src/model_vit.py
@@ -145,6 +145,7 @@ class ViTLitModule(L.LightningModule):
         assert embeddings.shape == torch.Size(
             [self.B, 65, 768]  # (batch_size, sequence_length, hidden_size)
         )
+        assert not torch.isnan(embeddings).any()  # ensure no NaNs in embedding
 
         # Save embeddings in npy format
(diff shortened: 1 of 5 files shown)


src/model_vit.py
```diff
@@ -126,6 +129,31 @@ def validation_step(self, batch: torch.Tensor, batch_idx: int) -> torch.Tensor:
...
+        # x: torch.Tensor = torch.randn(32, 13, 256, 256)  # BCHW
+
+        # Forward encoder
+        outputs_encoder: dict = self(x)
+
+        # Get embeddings generated from encoder
+        embeddings: torch.Tensor = outputs_encoder.last_hidden_state
+        assert embeddings.shape == torch.Size(
+            [self.B, 17, 768]  # (batch_size, sequence_length, hidden_size)
+        )
```
weiji14, line 146: @srmsoumya, if you have time, could you try to see why the outputs of the encoder have a shape like `(32, 17, 768)`, or `(batch_size, sequence_length, hidden_size)`? Specifically, I'm not sure what the sequence_length (size: 17) dimension is about, and couldn't quite figure it out from reading https://huggingface.co/docs/transformers/model_doc/vit_mae. More just something to understand, since we'll be moving to your MAE implementation in #47 later.

srmsoumya: @weiji14 we are masking out `75%` of the patches from the image. Given our image size is `256 x 256` and we use a patch size of `32`, that will give us `256 / 32` => `8 x 8` => `64 patches`. Masking out `75%` of these patches gives us `0.75 x 64 => 16 patches`. Adding 1 extra cls token gives us a total of `17 patches` to input into the transformer portion of the encoder. This results in a batch size of `batch_size: 32`  x `number of unmasked patches: 17` x `patch embedding: 768`. It is also the output we get from the encoder.

When creating the embeddings from the encoder of the model, we should switch-off the masking strategy.

weiji14: Wonderful explanation! Thanks @srmsoumya, I've disabled the masking in the `predict_step` now at commit f09d2e7f45c3d1f30ec7397e4874b26e70fce2a2, and the output is now (32, 65, 768) as expected. Here's a new sample embedding file: [embedding_0.npy.zip](https://github.com/Clay-foundation/model/files/13505741/embedding_0.npy.zip)

weiji14:
Gonna leave this for a day or so for review. There are a few nice-to-haves (e.g. better documentation on how to read the embeddings, a better filename that includes some spatiotemporal metadata, etc), but those can be done in follow-up PRs.

src/model_vit.py
weiji14: Hmm, do you have any papers describing when either the cls token or mean of the patch embeddings are used? Also @leothomas, do you have any insights on which method of collapsing the embeddings into a single 1d vector was used for the vector database projects you've worked on? Just want some extra context before decide on which way to go.

weiji14: Ok, after some internal discussion, we decided to go ahead with taking the mean along the spatial dim of the (8x8=64, 768) shaped tensor, to get a final shape of (1, 768). See commit 8eb179da631ffdf87bc4668322d7b13266e424d2, and the Excalidraw illustration in the top post.

brunosan: @danhammer you were asking me this before.

brunosan: I think you are referring to the mode?. I might be wrong, but anyway, it's most probably a small difference in most cases. 

I'm all in to store all 64 at least for one run, or a large sample so we can explore options when we get time. This would clarify then scattering within a tile, which I suspect will not matter for 90% of cases but will matter for strong edges like coast.

weiji14: > Just to clarify, the patches are are 32x32 (pixels) subsections of the overall image, correct? Why are we masking 75% of them? Is that due to cloud cover, or seeking patches only over land?

Yes, each patch are 32x32 pixels. We were masking 75% of the patches because that's how a Masked Autoencoder is trained (see previous PR at #37), but for inference/prediction, we shouldn't apply the mask.

> In regards to the `cls` token, it seems to be a [classification token](https://datascience.stackexchange.com/questions/66207/what-is-purpose-of-the-cls-token-and-why-is-its-encoding-output-important) - which represents the entire image.
> 
> The theory behind averaging the embeddings is that there is a very high liklyhood (but not a guarantee) that the [averages of 2 similar collections will be similar and the averages of 2 difference collections will be different](https://randorithms.com/2020/11/17/Adding-Embeddings.html). It seems that this has mostly been researched in the case of text embeddings, where the exact order of words may matter less than in the case of image patches. I suspect that averaging the embeddings for each patch may lose some of the physical relationships of the overall image, but would be very interesting to compare and contrast the two.
> 
> If we're going to use something like `pgvector`, we can easily have both types of embeddings colocated in the database and build a partial index for each!

That sounds good actually. We could save two files, a cls_token embedding (1x768), and a patch embedding (either 1x768, or 64x768)?

brunosan: If my intuition is correct, is not that surprising that most tiles are "semantically flat" since most tiles will be "on thing" all forest, or grass, or mountain ... The test for the need of richer embeddings would be to pick up examples with semantically rich, or even semantic polarization, like an image with both land and water, or city and forest. 

Also since we train with a MAE, I can see how each semantic patch will actually learn to include the expected semantics of the surrounding patches within the chip, not its own concent; so that the inference can recreate the missing bits, making the stdev within a patch smaller. 

Am I making sense?

weiji14: Yep, it would be good to explore what the embeddings look like for tiles that have more diverse land cover types. I've done just that by picking the most diverse tile we have, details at https://github.com/Clay-foundation/model/issues/35#issuecomment-1837027313

> Also since we train with a MAE, I can see how each semantic patch will actually learn to include the expected semantics of the surrounding patches within the chip, not its own concent; so that the inference can recreate the missing bits, making the stdev within a patch smaller.
> 
> Am I making sense?

Yes, the 1x768 embedding generated from each 32x32 patch actually contains information from the other patches. More details at https://github.com/Clay-foundation/model/discussions/67, and we can continue the discussion there!

weiji14:
Thanks for reviewing @srmsoumya, and everyone else for the comments. I'll merge this in now, noting that the code here generates (1, 768) shape averaged embeddings rather than the (1, 65, 768) shape raw embeddings. We can revise this later if we decide that we do want the raw embeddings.

PR #60, opened 2023-12-01
check for no data on a tile level in sentinel 1 vv and vh, sentinel 2 and DEM

lillythomas:
Adjusts the tiler function to check for nodata in not just Sentinel 2, but also Sentinel 1 VV and VH as well as the DEM.


yellowcap:
Ref https://github.com/Clay-foundation/model/issues/68

scripts/tile.py
```diff
@@ -37,8 +37,17 @@ def filter_clouds_nodata(tile):
     - bool: True if the tile is approved, False if rejected.
     """
     # Check for nodata pixels
-    nodata_pixel_count = int(tile.sel(band="B02").isin([NODATA]).sum())
-    if nodata_pixel_count:
+    bands_to_check = ["B02", "vv", "vh", "DEM"]
+    if int(tile.sel(band=bands_to_check[0]).isin([NODATA]).sum()):
+        print("Too much no-data")
+        return False
+    elif int(tile.sel(band=bands_to_check[1]).isin([NODATA]).sum()):
```
yellowcap, line 44: I think looking for `NODATA` values won't work because for `vv`, `vh`, and `DEM` we use `np.nan` for nodata values, not `0` like in Sentinel-2. In Sentinel-2, where we keep the data in Uint16, which does _not_ have a propoer `nan` value. But for the others its Float32, which does have nan value option. So the filter here should be something like

```python
int(np. isnan(tile.sel(band=bands_to_check[1])).sum())
```

scripts/tile.py
```diff
@@ -37,8 +37,17 @@ def filter_clouds_nodata(tile):
     - bool: True if the tile is approved, False if rejected.
     """
     # Check for nodata pixels
-    nodata_pixel_count = int(tile.sel(band="B02").isin([NODATA]).sum())
-    if nodata_pixel_count:
+    bands_to_check = ["B02", "vv", "vh", "DEM"]
+    if int(tile.sel(band=bands_to_check[0]).isin([NODATA]).sum()):
```
yellowcap, line 41: Since this is list based, we could consider using a loop here.

yellowcap (changes requested):
The nodata filter for float 32 needs to be slightly different.

yellowcap:
Rebased from main, and added small fix on the routine (`"DEM"` band is named lowercase `"dem"` in the tile xarray)

PR #63, opened 2023-12-03
Setting the model license to OpenRail-M

brunosan:
This PR sets the license for Clay the model, which includes the embeddings, model file, and also propagates to all finetune applications or derivatives of embeddings.

As we said on #1 we are open source and open model.
 
Open Source has established standards that make it friction-free for downstream users to adopt the code we make. Hence [we choose Apache](https://github.com/Clay-foundation/model/blob/main/LICENSE), one of the most permissive licenses. There are new ML source licenses, but I'm not convinced there is enough value to adopt these less known options versus something much more widely known and adopted like Apache.

For Open models, the option that seems most permissive and standard, as identified by @weiji14, is [OpenRail-M](https://www.licenses.ai/ai-pubs-open-rails-vz1). 

This PR sets the OpenRail-M as the license for our model. Some of the use restrictions seem meant for LLMs, but I rather change the minimum from the template, hence there are no modifications, expect the 3 lines at the top stating our name.

### DIFF
--- a/model-license.md
+++ b/model-license.md
@@ -1,0 +1,119 @@
+CLAY OpenRAIL-M LICENSE
+This OpenRAIL-M License is applicable by default to all trained Clay Models.
+
+From:
+AI PUBS OPEN RAIL-M LICENSE
+Version 0.1, March 2, 2023
+
+http://licenses.ai/
+
+  Section  I:  PREAMBLE
+
+This OpenRAIL-M License is generally applicable to any machine-learning Model.
+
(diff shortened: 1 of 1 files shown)


yellowcap (approved):
That is great, the only comment/question I would have is if this is the right way to place the license. I.e. is the starndard to have a `LICENSE` file that references the code, and a `model-license` file to specify for the model? Just wondering how people know what refers to what.

weiji14:
Not sure what the standard is, but was wondering if renaming the file to LICENSE-MODEL.md (so that it is just below the current LICENSE) would be good?

PR #66, opened 2023-12-04
Let LightningDataModule return spatiotemporal metadata

weiji14:
Making the LightningDataModule return not only the image, but also spatiotemporal metadata such as:

- bounding box (bbox): (xmin, ymin, xmax, ymax)
- coordinate reference system (epsg): EPSG code
- date: YYYY-MM-DD string format

Note:
- The bbox and epsg is in the raster image's native UTM projection for now, rather than lonlat coordinates.

This is part 1/2 of adding spatiotemporal metadata to the output embedding table later, as mentioned at https://github.com/Clay-foundation/model/issues/35#issuecomment-1839738582.

### DIFF
--- a/src/datamodule.py
+++ b/src/datamodule.py
@@ -12,10 +12,25 @@ import torchdata
 
 
 # %%
-def _array_to_torch(filepath: str) -> torch.Tensor:
+def _array_to_torch(filepath: str) -> dict[str, torch.Tensor | str]:
     """
-    Read a GeoTIFF file using rasterio into a numpy.ndarray, and convert it
-    to a torch.Tensor (float16 dtype).
+    Read a GeoTIFF file using rasterio into a numpy.ndarray, convert it to a
+    torch.Tensor (float16 dtype), and also output spatiotemporal metadata
+    associated with the image.
+
+    Parameters
+    ----------
+    filepath : str
+        The path to the GeoTIFF file.
+
+    Returns
+    -------
+    outputs : dict
+        A dictionary containing the following items:
+        - image: torch.Tensor - multi-band raster image with shape (Band, Height, Width)
+        - bbox: torch.Tensor - spatial bounding box as (xmin, ymin, xmax, ymax)
+        - crs: torch.Tensor - coordinate reference system as an EPSG code
+        - date: str - the date the image was acquired in YYYY-MM-DD format
     """
     # GeoTIFF - Rasterio
     with rasterio.open(fp=filepath) as dataset:
--- a/src/datamodule.py
+++ b/src/datamodule.py
@@ -2,6 +2,8 @@
 LightningDataModule to load Earth Observation data from GeoTIFF files using
 rasterio.
 """
+import pathlib
+
 import lightning as L
 import numpy as np
 import rasterio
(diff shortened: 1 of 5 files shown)


src/datamodule.py
```diff
@@ -17,10 +19,20 @@ def _array_to_torch(filepath: str) -> torch.Tensor:
...
 
-    return tensor
+        # Get spatial bounding box and coordinate reference system in UTM projection
+        bbox: torch.Tensor = torch.as_tensor(  # xmin, ymin, xmax, ymax
+            data=dataset.bounds, dtype=torch.float64
+        )
+        crs: int = torch.as_tensor(data=dataset.crs.to_epsg(), dtype=torch.int32)
+
+        # Get date
+        date: str = pathlib.Path(filepath).name[15:25]  # YYYY-MM-DD format
```
weiji14, line 33: Don't like that we have to parse the date from the filename. @yellowcap, I hinted on this at https://github.com/Clay-foundation/model/pull/54#discussion_r1413281596, but would it be possible to save the datetime information in the GeoTIFF's metadata somehow?

src/datamodule.py
```diff
@@ -10,17 +12,42 @@
...
+
+    return {"image": tensor, "bbox": bbox, "crs": crs, "date": date}
```
weiji14, line 50: Tempted to store this in an Arrow Table with [`FixedShapedTensorArray`](https://arrow.apache.org/docs/python/generated/pyarrow.FixedShapeTensorArray.html) for the image and bbox 'columns'. Possibly revisit this in the future.

weiji14:
Gonna leave this up for review for a day or so before merging. Once merged, I'll proceed to work on part 2/2, which is to get the model to output embeddings with spatiotemporal metadata columns!

src/datamodule.py
yellowcap: After the next pipeline run you'll be able to get the date using
```python
dataset.tags()["date"]
```

src/datamodule.py
```diff
@@ -10,17 +12,42 @@
...
+        # Get image data
         array: np.ndarray = dataset.read()
         tensor: torch.Tensor = torch.as_tensor(data=array.astype(dtype="float16"))
 
-    return tensor
+        # Get spatial bounding box and coordinate reference system in UTM projection
+        bbox: torch.Tensor = torch.as_tensor(  # xmin, ymin, xmax, ymax
+            data=dataset.bounds, dtype=torch.float64
+        )
+        crs: int = torch.as_tensor(data=dataset.crs.to_epsg(), dtype=torch.int32)
```
yellowcap, line 45: This variable could be called `epsg` to make clear its a crs as epsg integer.

yellowcap (approved):
Looks good to me. Just one small comment on variable name.

src/datamodule.py
weiji14: Was considering this actually, I've changed it in 6a0e8cfe14ec4a1a8ffddf4e54dd8ffaea41e4be!

src/datamodule.py
weiji14: Awesome, can't wait!

Issue #78, opened 2023-12-07
Increase tile size to 512x512 pixels


yellowcap:
For easier use in benchmark applications, we should increase the tile size to 512 pixels squared. This does not have large implications for the model training and can help with benchmarking. It will also reduce the number of files to be stored.

mattpaul:
Hi @yellowcap, just looking for some clarity - can you confirm that:
- does the change to `TILE_SIZE` impact embeddings as well?
- is "tile" here synonymous with "chip" in this discussion https://github.com/Clay-foundation/model/discussions/67 ?

if so, @weiji14 is it safe to assume that:
- patch size of 32x32 is unchanged,

and therefore we now expect:
- 8x8x4 = 256 patches per chip
- embeddings per chip file will now be tensors of shape (1, 256+1, 768)

weiji14:
The patch size might change from 32x32 to 64x64 (to be decided). We are still serving a single embedding of size 1x768 for each 512x512 chip, which is the mean/average taken along the spatial dimension, see discussion at https://github.com/Clay-foundation/model/pull/56#discussion_r1411249922.

The (1, 256+1, 768) embedding was the unaveraged embedding, that I first shared initially to see the in-chip distribution of the embeddibgs, sorry if that was confusing.

yellowcap:
@mattpaul the tiles here are indeed the "chips". The 512 size is just for collection to match better against benchmark datasets. For the model we will most likely still feed it 256x256 chips. In any case the output embedding size is a hyperparameter of the model, which is relatively independent from the input size.

mattpaul:
@yellowcap thanks for the clarification. I guess I was assuming that whenever we use the term "chip" it would mean an image of the same size regardless of context but that's not necessarily always the case. Good to know!

I gather there are benchmark datasets out there that we would like to compare against or otherwise work with, and they typically have chips of size 512x512, therefore we're modifying our chip factory to process MSRG tiles into chips of similar size for our training dataset of images.

When you say "for the model we will most likely still feed it 256x256 chips," this suggests that there is an additional model input transform that occurs between the chip factory output of 512x512 and the model input layer which will likely continue to expect input of 256x256 = 65535 pixels, correct? Perhaps something as simple as consuming _one_ 512x512 chip and producing _four_ 256x256 chips with which to feed the model.

[...]

PR #85, opened 2023-12-11
Adapt model to load 512x512 images from s3 bucket

weiji14:
Modify the ViT MAE model to accept input images of size 512x512 pixel after #78. Also making a few small enhancements to the datapipe.

TODO:
 - [x] Change hardcoded image_size from 256 to 512, patch_size from 32 to 64.
 - [x] Allow loading of GeoTIFF files directly from s3 bucket instead of local drive
 - [x] Optimize data loading using sharding filter
 - [ ] etc

### DIFF
--- a/src/datamodule.py
+++ b/src/datamodule.py
@@ -117,19 +117,22 @@ class GeoTIFFDataPipeModule(L.LightningDataModule):
 
             # Step 3 - Read GeoTIFF into numpy array, batch and convert to torch.Tensor
             self.datapipe_train = (
-                dp_train.map(fn=_array_to_torch)
+                dp_train.sharding_filter()
+                .map(fn=_array_to_torch)
                 .batch(batch_size=self.batch_size)
                 .collate()
             )
             self.datapipe_val = (
-                dp_val.map(fn=_array_to_torch)
+                dp_val.sharding_filter()
+                .map(fn=_array_to_torch)
                 .batch(batch_size=self.batch_size)
                 .collate()
             )
 
         elif stage == "predict":  # prediction loop
             self.datapipe_predict = (
-                self.dp_paths.map(fn=_array_to_torch)
+                self.dp_paths.sharding_filter()
+                .map(fn=_array_to_torch)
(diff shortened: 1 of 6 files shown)


src/datamodule.py
```diff
@@ -87,38 +87,51 @@ def __init__(
...
-            root=self.data_path, masks="*.tif", recursive=True, length=423
-        )
+        # Step 1 - Get list of GeoTIFF filepaths from s3 bucket or data/ folder
+        if self.data_path.startswith("s3://"):
+            dp = torchdata.datapipes.iter.IterableWrapper(iterable=[self.data_path])
+            self.dp_paths = dp.list_files_by_s3(masks="*.tif")
+        else:  # if self.data_path is a local data path
+            self.dp_paths = torchdata.datapipes.iter.FileLister(
+                root=self.data_path, masks="*.tif", recursive=True
+            )
```
weiji14, line 108: Note that loading data from s3 is still a bit slower than a local data folder, but it can be useful for cases like one-off inference/prediction when we don't want to store large amounts of data locally.

weiji14:
Gonna merge this PR directly, as it consists of mostly minor tweaks (which I've accumulated over the past few weeks).

src/datamodule.py
```diff
@@ -87,38 +87,51 @@ def __init__(
...
+            Whether to setup the datapipe for the training/validation loop, or
+            the prediction loop. Choose from either 'fit' or 'predict'.
         """
-        # Step 1 - Get list of GeoTIFF filepaths from data/ folder
-        dp_paths = torchdata.datapipes.iter.FileLister(
-            root=self.data_path, masks="*.tif", recursive=True, length=423
-        )
+        # Step 1 - Get list of GeoTIFF filepaths from s3 bucket or data/ folder
+        if self.data_path.startswith("s3://"):
+            dp = torchdata.datapipes.iter.IterableWrapper(iterable=[self.data_path])
```
robmarkcole, line 103: Noting datapipes are deprecated, is this a long term solution?

PR #88, opened 2023-12-13
Callback function to log Masked Autoencoder reconstructions to WandB

weiji14:
To visually inspect how the Masked Autoencoder is performing over the training run in terms of reconstructing to original image.

Implemented as a [Lightning Callback](https://lightning.ai/docs/pytorch/2.1.0/extensions/callbacks.html) function which runs at the end of a validation loop's mini-batch ([`on_validation_batch_end`](https://lightning.ai/docs/pytorch/2.1.0/extensions/callbacks.html#on-validation-batch-end)). A sample of 6 image pairs (original + reconstructed, so total of 12) are uploaded online.

Example usage with [LightningCLI](https://lightning.ai/docs/pytorch/stable/cli/lightning_cli_advanced_3.html#trainer-callbacks-and-arguments-with-class-type):

Samples of histogram equalized RGB images, and the reconstructed outputs (only random noise as this is early in the training).

image

```bash
python trainer.py fit --trainer.max_epochs=2 \
                      --data.data_path=data/32VLM \
                      --data.num_workers=4 \
                      --trainer.logger=WandbLogger \
                      --trainer.logger.project=clay \
                      --trainer.logger.save_dir=checkpoints \
                      --trainer.callbacks+=LogMAEReconstruction \
```

TODO:
- [x] Add wandb dependency
- [x] Initial implementation to upload RGB images
- [x] Do proper histogram equalization of the images
- [x] Add a unit test

TODO in the future:
- [ ] Upload SAR and DEM images

References:
- https://lightning.ai/docs/pytorch/2.1.0/common/trainer.html#callbacks
- https://wandb.ai/wandb/wandb-lightning/reports/Image-Classification-using-PyTorch-Lightning--VmlldzoyODk1NzY
- https://github.com/ashleve/lightning-hydra-template/blob/wandb-callbacks/src/callbacks/wandb_callbacks.py#L245
- https://gitlab.com/frontierdevelopmentlab/2022-us-sarchangedetection/deepslide/-/blob/main/src/callbacks/wandb_callbacks.py?ref_type=heads

### DIFF
--- a/src/tests/test_callbacks.py
+++ b/src/tests/test_callbacks.py
@@ -16,12 +16,13 @@ from src.model_vit import ViTLitModule
 
 
 # %%
-def test_callbacks_wandb_log_mae_reconstruction():
+def test_callbacks_wandb_log_mae_reconstruction(monkeypatch):
     """
     Ensure that the LogMAEReconstruction callback can log a set of images to
     Weights & Biases.
     """
     wandb = pytest.importorskip(modname="wandb")
+    monkeypatch.setenv("WANDB_CONSOLE", "off")
 
     with tempfile.TemporaryDirectory() as tmpdirname:
         pl_module: L.LightningModule = ViTLitModule()
--- a/src/tests/test_callbacks.py
+++ b/src/tests/test_callbacks.py
@@ -57,6 +57,8 @@ def test_callbacks_wandb_log_mae_reconstruction(monkeypatch):
             ]
         )
 
+        del trainer, pl_module
+
     # Check that images logged by WandB have the correct caption and format
     assert len(wandb_images) == 8  # noqa: PLR2004
     assert all(isinstance(w, wandb.Image) for w in wandb_images)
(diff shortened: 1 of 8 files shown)


src/tests/test_callbacks.py
```diff
@@ -0,0 +1,66 @@
...
+        # Check that wandb saved some log files to the temporary directory
+        # assert os.path.exists(path := f"{tmpdirname}/wandb/latest-run/")
+        # assert set(os.listdir(path=path)) == set(
+        #     [
+        #         f"run-{trainer.logger.version}.wandb",
+        #         "tmp",
+        #         "files",
+        #         "logs",
+        #     ]
+        # )
```
weiji14, line 58: Setting `WANDB_MODE="disabled"` in this unit test and commenting these lines that check for log files being saved, because GitHub Actions keeps failing with an error like `Error: Process completed with exit code 255`, even though all the unit tests pass on pytest :confused: The unit test does work locally though when I uncomment this block and use `WANDB_MODE="offline"`, so not sure what's going on.

weiji14 (commented):
Sunk wayyy too much time trying to debug why the unit test was failing on GitHub Actions but not locally (see below), so didn't get to do the SAR and DEM plots. Will implement those in a follow-up PR instead.

weiji14:
Gonna merge this in first and combine with the other wandb callback being developed at #47.

Issue #70, opened 2023-12-06
Timestamp in TIF metadata for data pipeline output


yellowcap:
To facilitate its use and persist beyond the file name.

See also https://github.com/Clay-foundation/model/pull/54#pullrequestreview-1761464883

weiji14:
Looking at whether there's a standard to encode the acquisition datetime in a GeoTIFF at https://gdal.org/drivers/raster/gtiff.html#metadata. A couple of options:

- [`ACQUISITIONDATETIME`](https://gdal.org/user/raster_data_model.html#imagery-domain-remote-sensing)
- [`TIFFTAG_DATETIME`](https://gdal.org/user/raster_data_model.html#metadata)

Both of these expect a datetime (with year, month, date, hour, minute second) though, but currently we are only storing YYYY-MM-DD. Which one should we go for?

yellowcap:
> Looking at whether there's a standard to encode the acquisition datetime in a GeoTIFF at https://gdal.org/drivers/raster/gtiff.html#metadata. A couple of options:

Sorry only saw this now. Went with simple `date`. But can adapt in v3.

yellowcap:
Looking into this, changing this tag needs refactoring also in the data-pipeline, finding the simple "date" tag and replace it with the choice of use [ACQUISITIONDATETIME](https://gdal.org/user/raster_data_model.html#imagery-domain-remote-sensing) or [TIFFTAG_DATETIME](https://gdal.org/user/raster_data_model.html#metadata).

yellowcap:
This is in the[ `date` tag in the tiff files for v0.1 and v0.2](https://github.com/Clay-foundation/model/blob/8f1d8692794e54c27b3a9ca9f6f59c8746819fa9/src/datamodule.py#L63), and will move into the STAC Item / stacchip index for v1.

PR #91, opened 2023-12-19
Let ClayDataModule return same spatiotemporal fields as GeoTIFFDataModule

weiji14:
Align the torch Dataset-based ClayDataModule developed in #47 with the torchdata-based GeoTIFFDataModule (mainly #66), so that both are consistent in terms of the input parameters and returned values.

TODO:
- [x] Standardize on the parameter `data_dir` with a `str` type
- [x] Get YYYY-MM-DD timestamp from GeoTIFF's metadata rather than filename, xref #70
- [x] Simplify calculation of latlon centroid value
- [x] Let ClayDataset return the UTM bounding box and EPSG code as in #66

This is part 1/3 of setting things up for the next round of embedding generation to conform to the GeoParquet table format in #86.

### DIFF
--- a/src/datamodule.py
+++ b/src/datamodule.py
@@ -56,8 +56,8 @@ class ClayDataset(Dataset):
         chip = rasterio.open(chip_path)
 
         # read timestep & normalize
-        ts = chip_path.parent.name
-        year, month, day = self.normalize_timestamp(ts)
+        date = chip.tags()["date"]  # YYYY-MM-DD
+        year, month, day = self.normalize_timestamp(date)
 
         # read lat,lon from UTM to WGS84 & normalize
         bounds = chip.bounds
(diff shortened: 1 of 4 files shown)


src/datamodule.py
```diff
@@ -80,6 +81,9 @@ def __getitem__(self, idx):
 
         # remove nans and convert to tensor
         cube["pixels"] = torch.nan_to_num(torch.as_tensor(data=cube["pixels"]), nan=0.0)
+        cube["bbox"] = torch.as_tensor(data=cube["bbox"], dtype=torch.float64)
+        cube["epsg"] = torch.as_tensor(data=cube["epsg"], dtype=torch.int32)
+        cube["date"] = str(cube["date"])
         cube["latlon"] = torch.as_tensor(data=cube["latlon"])
         cube["timestep"] = torch.as_tensor(data=cube["timestep"])
         cube["path"] = str(chip_path.absolute())
```
weiji14, line 89: Still need this rename `cube["path"]` in ClayDataModule to match with GeoTIFFDataModule's `cube["source_url"]`, but will do that in a follow-up PR when I handle listing of GeoTIFF files from s3 buckets.

Issue #68, opened 2023-12-06
Catch Nodata for S1 and DEM


yellowcap:
See discussion in https://github.com/Clay-foundation/model/pull/54#pullrequestreview-1754125092

PR #92, opened 2023-12-19
Allow ClayDataModule to load GeoTIFF files directly from s3

weiji14:
Similar to work done in #85 on the GeoTIFFDataPipeModule, this PR implements similar functionality in ClayDataModule to load GeoTIFF files from an s3 bucket. Plus a few more minor tweaks to align both LightningDataModules.

Implementation uses `torchdata`'s [S3FileLister](https://pytorch.org/data/0.7/generated/torchdata.datapipes.iter.S3FileLister.html#torchdata.datapipes.iter.S3FileLister) to get the files, but instead of returning an iterator, a list is returned.

TODO:
- [x] Allow ClayDataModule to load GeoTIFF files directly from s3
- [x] Rename `datacube["path"]` to `datacube["source_url"]` to match #86
- [x] Implement predict dataloader
- [x] Add a unit test

Continuing on from #91, this PR is part 2/3 of working towards generating new embeddings from the model developed at #47.

### DIFF
--- a/src/datamodule.py
+++ b/src/datamodule.py
@@ -80,7 +80,7 @@ class ClayDataset(Dataset):
         cube = self.read_chip(chip_path)
 
         # remove nans and convert to tensor
-        cube["pixels"] = torch.nan_to_num(torch.as_tensor(data=cube["pixels"]), nan=0.0)
+        cube["pixels"] = torch.as_tensor(data=cube["pixels"], dtype=torch.float16)
         cube["bbox"] = torch.as_tensor(data=cube["bbox"], dtype=torch.float64)
         cube["epsg"] = torch.as_tensor(data=cube["epsg"], dtype=torch.int32)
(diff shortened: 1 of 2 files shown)


src/datamodule.py
```diff
@@ -80,13 +80,16 @@ def __getitem__(self, idx):
         cube = self.read_chip(chip_path)
 
         # remove nans and convert to tensor
-        cube["pixels"] = torch.nan_to_num(torch.as_tensor(data=cube["pixels"]), nan=0.0)
+        cube["pixels"] = torch.as_tensor(data=cube["pixels"], dtype=torch.float16)
```
weiji14, line 87: Heads up @srmsoumya that I've remove the NaN to 0 clipping here, since the new batch of GeoTIFF files shouldn't have NaNs anymore per https://github.com/Clay-foundation/model/issues/68.

PR #95, opened 2023-12-20
Refactor model for multi-device usage and easier disabling of masking

weiji14:
A couple of small changes to `model_clay.py` after initial implementation at #47.

- [x] Handle layers across cpu/gpu devices a bit better, to prevent errors like `RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!`
- [x] Let the masking in the Masked Autoencoder be dynamic, so that setting `self.model.encoder.mask_ratio = 0` would disable the masking.

First one is cherry-picked from @srmsoumya's `ddp` branch. Second one will be helpful for the embedding generation later on.

### DIFF
--- a/src/model_clay.py
+++ b/src/model_clay.py
@@ -68,9 +68,6 @@ class Encoder(nn.Module):
         self.num_spatial_patches = (image_size // patch_size) ** 2
         self.num_group_patches = len(band_groups)
         self.num_patches = self.num_spatial_patches * self.num_group_patches
-        self.num_masked_patches = int(
-            self.mask_ratio * self.num_patches
-        )  # Number of patches to be masked out
 
         # Split the embedding dimensions between spatial & band patches equally
         pos_dim = band_dim = dim // 2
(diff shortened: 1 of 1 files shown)


src/model_clay.py
```diff
@@ -281,6 +279,9 @@ def forward(self, datacube):
             cube
         )  # [B G L D] - patchify & create embeddings per patch
 
+        # Move position & band encoding to the device
+        self.pos_encoding = self.pos_encoding.to(patches.device)
+        self.band_encoding = self.band_encoding.to(patches.device)
```
weiji14, line 284: @srmsoumya, I feel like we can also remove the `.to(device)` call here?
```suggestion
```

src/model_clay.py
```diff
@@ -506,6 +505,10 @@ def forward(self, encoded_unmasked_patches, unmasked_indices, masked_indices):
             encoded_unmasked_patches[:, -2:, :],
         )  # [B (GL:(1 - mask_ratio)) D], [B 2 D]
 
+        # move position & band encoding to the device
+        self.pos_encoding = self.pos_encoding.to(encoded_unmasked_patches.device)
+        self.band_encoding = self.band_encoding.to(encoded_unmasked_patches.device)
+
```
weiji14, line 511: Same here, this  `.to(device)` call can probably be removed.
```suggestion
```

src/model_clay.py
weiji14: Ah, I think I found the core issue. The `pos_encoding` and `band_encoding` tensors are declared in the `__init__` method, before Lightning has managed to cast them to the correct device.

https://github.com/Clay-foundation/model/blob/ee74c91b99cbf9a7c304c0d37335c128d1ae7566/src/model_clay.py#L89-L95

Ideally, we'll need to define the device at this point. Let me see if there's a way to do that.

weiji14: Ok, went with using `self.register_buffers` at commit 25f83e62863da5e5ba4a0a8625a8ef8d1b2afd2a. Have set `persistent=False` since those pos_encoding and band_encoding tensors are fixed, and won't need to be saved to the `state_dict`.

weiji14:
Ok, gonna merge this patch in directly since others may benefit from the improve cuda/cpu device handling. Plus I'll need these patches for the embedding factory script.

PR #102, opened 2023-12-26
Nbdev for Clay Documentation, Clay SDK and Clay notebooks all at once

brunosan:
I've started to work on documentation. I decided to start over using [nbdev](https://nbdev.fast.ai/)

It's a bit convoluted and I'm over my head coding here, but it seems highly attractive to do all this all at once without extra work:
* Explore the model with notebooks anyone can run.
* All functions and Classes created can be `pip install`ed, making it a small SDK for Clay.
* Notebooks help generate documentation website for all functions, with links to the code 
* Documentation and sample code tests run on CI.

@weiji14 your work is still in the history, and we can revert to that. I've migrated your text to this PR.

It's early, but I wanted to give a heads up in case others have experience on this.

GH Action is already rendering a draft site on [https://clay-foundation.github.io/model/](https://clay-foundation.github.io/model/)

I've also focused on the v0 release notes, but we still miss a lot of information. WIP.

image

PR #103, opened 2023-12-27
Docs/SDK/notebooks with nbdev [try 2]

brunosan:
This take into account some of @weiji14 comments on #102 (which I abandoned to not deal with a polluted git history).

I will not spend much time on the architecture (nbdev vs other options). I do like the option to have all 3 assets at once (documentation htmls, SDK and notebooks) but I understand `nbdev` might be too rare on stale.

I'll focus meanwhile on the notebooks themselves and we can port the content to other tools.

In the end we need documentation asap.


brunosan:
After thinking about it, I'll spin out this folder with documentation to a new repository.
I think that our goal to create a good comprehensive documentation, templates, ... will bloat this repo with significantly functionally different content.

Spinning out on https://github.com/Clay-foundation/documentation/pull/1
I'll clean up the publish workflow on a PR next

PR #104, opened 2023-12-28
Remove docs, spun out

brunosan:
We are spinning out the documentation to a separate repo. This cleans up the workflow.

See https://github.com/Clay-foundation/model/pull/103#issuecomment-1871277647


yellowcap:
Deferring this to later, to keep things simple for now we decided to keep the internal documentation. We can split this out into separate repo easily in the future.

brunosan:
Agree. Also unpublished the other documentation repo, and deleted the associated pipy.
I'll replace the text on Hugging face with the release notes once we have them.

PR #109, opened 2024-01-04
Add multigpu support & UI to test embeddings 

srmsoumya:
- Update MEAN & STD for 10% of the data used for CLAY v0 model training
- Add streamlit UI with support for vector search & arithmetic
- Fix issue with tensors on different cuda devices when doing multi-gpu training


clayground.py
weiji14, line 1: Is this clayground.py script meant to be part of the model repo?

trainer.py
```diff
@@ -30,28 +31,30 @@ def cli_main(
     save_config_callback=None,
     seed_everything_default=42,
     trainer_defaults: dict = {
-        "accelerator": "auto",
+        "accelerator": "gpu",
```
weiji14, line 34: Prefer to set `accelerator="auto"` to allow this `trainer.py` to work on the CPU out of the box too.
```suggestion
        "accelerator": "auto",
```

src/model_clay.py
```diff
@@ -287,6 +289,9 @@ def forward(self, datacube):
             cube
         )  # [B G L D] - patchify & create embeddings per patch
 
+        # Move position & band encoding to the device
+        self.pos_encoding = self.pos_encoding.to(patches.device)
+        self.band_encoding = self.band_encoding.to(patches.device)
```
weiji14, line 294: Hmm, I thought these `.to()` lines aren't needed anymore after https://github.com/Clay-foundation/model/pull/95#discussion_r1432381686? Are you still getting errors like `RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu`?

trainer.py
```diff
@@ -30,28 +31,30 @@ def cli_main(
...
                 save_weights_only=True,
                 verbose=True,
             ),
-            # LearningRateMonitor(logging_interval="step"),
-            # LogIntermediatePredictions(),
+            LearningRateMonitor(logging_interval="step"),
+            LogIntermediatePredictions(),
         ],
-        "logger": False,  # WandbLogger(project="CLAY-v0", log_model=False)
+        "logger": [WandbLogger(project="CLAY-v0", log_model=False)],
```
weiji14, line 56: Ok to add these loggers for now, but I'd like to disable them by default so that new users don't need to install and setup `wandb` and such on the first run. We could potentially use LightningCLI's YAML config files - https://lightning.ai/docs/pytorch/2.0.2/cli/lightning_cli_advanced.html to store some of these advanced configurations.

src/model_clay.py
```diff
@@ -773,15 +772,41 @@ def clay_large(**kwargs):
...
     }
-    model = CLAY(**args.update(kwargs))
+    args.update(kwargs)
+    model = CLAY(**args)
     return model
 
 
 class CLAYModule(L.LightningModule):
-    def __init__(self, lr=1e-4, wd=0.05, b1=0.9, b2=0.95):
+    def __init__(
```
weiji14, line 781: Add this to silence the [PLR0913 Too many arguments](https://docs.astral.sh/ruff/rules/too-many-arguments) in function definition (9 > 6) warning
```suggestion
    def __init__(  # noqa: PLR0913
```

weiji14 (commented):
Happy New Year @srmsoumya! Just some comments to help you pass the CI tests and linter warnings, plus a few optional recommendations/notes.

clayground.py
srmsoumya: It can stay in the model repository for now, will be useful to play around & test models capabilities.
We can plan to move this to the vector db repo later.

trainer.py
srmsoumya: Yes, done!

trainer.py
srmsoumya: Agreed

src/model_clay.py
srmsoumya: Thanks!

clayground.py
```diff
@@ -0,0 +1,130 @@
...
+        {"tile": "49KHR", "idx": "0020", "year": 2017},
+        {"tile": "55LBC", "idx": "0075", "year": 2022},
+    ]
+
+    filter = " OR ".join(
+        [
+            f"(tile == '{chip['tile']}' AND idx == '{chip['idx']}') AND year == {chip['year']}"
+            for chip in chips
+        ]
+    )
```
weiji14, line 30: Just need to wrap L27 to <88 characters to make the linter happy.

weiji14 (approved):
Cool, just one more thing to fix and should be good to merge!

clayground.py
weiji14: By the way @srmsoumya, [`filter`](https://docs.python.org/3/library/functions.html#filter) is a built-in Python function. Can we change the variable name to something else?

clayground.py
```diff
@@ -0,0 +1,130 @@
...
+    result = tbl.search().where(filter, prefilter=True).to_pandas()
+    return result
+
+
+# Load embeddings
+@st.cache_resource()
+def connect_to_database():
+    db = lancedb.connect("nbs/embeddings")
+    tbl = db.open_table("clay-v001")
```
weiji14, line 39: Will need to document how the `nbs/embeddings' LanceDB vector database is created. Can do this afterwards.

clayground.py
weiji14: Ok, I just went ahead and renamed `filter` to `tile_filter` at commit f4731c63186723a26563364de9994f42a08ee6a7

trainer.py
weiji14: Another reason is that turning on the WandbLogger by default here also means `wandb` is invoked during inference/prediction, and we would need to do something like `WANDB_MODE=disabled python trainer.py predict ...` to not use wandb at all.

src/model_clay.py
weiji14: There were two more lines that were missed, so I removed them in 52243a9a6c96b8b7342ad711cd3211ce5c9ccb80

PR #116, opened 2024-01-12
Clay over aoi

yellowcap:
Updated pipeline and added notebook to docs to run clay over custom aois over custom date ranges.

I tested the script locally extensively and it works, having more people trying it won't hurt to see if there are any things that are missing / that I did not catch in my local runs.


brunosan:
I merged this PR into #118. Which would close this one when merged.

docs/clay_over_aoi.ipynb
```diff
@@ -0,0 +1,727 @@
...
+    "            },\n",
+    "            \"geometry\": {\n",
+    "                \"type\": \"Polygon\",\n",
+    "                \"coordinates\": [ [ [ 85.050328768992927, 19.494998302498729 ], [ 85.169749958884921, 19.441346370958311 ], [ 85.968028516820738, 19.799945705366934 ], [ 86.029742823006544, 20.15322442052609 ], [ 86.104280881127053, 20.516965603502392 ], [ 85.404584916189307, 20.564248936237991 ], [ 84.945334300027469, 20.339711117597215 ], [ 84.816295296184407, 19.799945705366934 ], [ 85.050328768992927, 19.494998302498729 ] ] ]\n",
+    "            }\n",
+    "        }\n",
+    "    ]\n",
+    "}\"\"\"\n",
+    "aoi = gpd.read_file(aoi_src, driver=\"GeoJSON\")\n",
+    "aoi.geometry[0]"
```
weiji14, line 267: Can we show how to specify the AOI via [`shapely.Point`](https://shapely.readthedocs.io/en/2.0.2/reference/shapely.Point.html) or [`shapely.box`](https://shapely.readthedocs.io/en/2.0.2/reference/shapely.box.html) instead? GeoJSON is a little harder to construct manually.

docs/clay_over_aoi.ipynb
```diff
@@ -0,0 +1,727 @@
...
+      "Saved 2 rows of embeddings of shape (768,) to /home/tam/Documents/repos/model/data/embeddings/44QRH_20200328_20200328_v001.gpq\n",
+      "Saved 2 rows of embeddings of shape (768,) to /home/tam/Documents/repos/model/data/embeddings/45QTC_20200328_20200328_v001.gpq\n",
+      "Predicting DataLoader 0: 100%|██████████████████| 14/14 [00:02<00:00,  6.77it/s]\n",
+      "Done!\n"
+     ]
+    }
+   ],
+   "source": [
+    "! wandb disabled\n",
+    "! python trainer.py predict --ckpt_path=data/checkpoints/Clay_v0.1_epoch-24_val-loss-0.46.ckpt \\\n",
```
weiji14, line 690: Yeah, we should look to disable wandb by default in `trainer.py` as mentioned at https://github.com/Clay-foundation/model/pull/109#discussion_r1444170134. Relevant lines are here:

https://github.com/Clay-foundation/model/blob/0145e55bcf6bd3e9b19f5c07819a1398b6a22c35/trainer.py#L53-L56

scripts/datacube.py
```diff
@@ -423,28 +422,54 @@ def convert_attrs_and_coords_objects_to_str(data):
...
-    help="For debugging, subset x and y to this pixel window.",
+    help="If specified, this path will be used to write the tiles locally"
+    "Otherwise a temp dir will be used.",
+)
+@click.option(
+    "--dateranges",
+    required=False,
+    default="",
+    type=str,
+    help="Comma separated list of date ranges, each provided as yy-mm-dd/yy-mm-dd.",
```
weiji14, line 442: Is it year as 4 numbers (e.g. `2024`) or as 2 numbers (e.g. `24`)?

```suggestion
    help="Comma separated list of date ranges, each provided as yyyy-mm-dd/yyyy-mm-dd.",
```

scripts/datacube.py
```diff
@@ -423,28 +422,54 @@ def convert_attrs_and_coords_objects_to_str(data):
 @click.option(
     "--subset",
     required=False,
+    default="",
+    help="For debugging, subset x and y to this pixel window as a comma"
+    "separated string of 4 integers.",
```
weiji14, line 427: I see that this is for debugging only, but just to clarify, this option is for cropping the tiles/chips to a specified height and width?

scripts/tile.py
```diff
@@ -66,67 +66,77 @@ def tiler(stack, date, mgrs, bucket):
...
+                mgrs=mgrs,
+                date=date.replace("-", ""),
+                version=VERSION,
+                counter=str(counter).zfill(4),
+            )
+            tile.rio.to_raster(name, compress="deflate")
+
+            with rasterio.open(name, "r+") as rst:
+                rst.colorinterp = color
+                rst.update_tags(date=date)
```
weiji14, line 119: Do you want to handle https://github.com/Clay-foundation/model/issues/70 here, i.e. use [ACQUISITIONDATETIME](https://gdal.org/user/raster_data_model.html#imagery-domain-remote-sensing) or [TIFFTAG_DATETIME](https://gdal.org/user/raster_data_model.html#metadata)? Or do that in a separate PR?

docs/clay_over_aoi.ipynb
```diff
@@ -0,0 +1,727 @@
...
+      "\n",
+      "2024-01-12 16:21:57 (3.04 MB/s) - ‘data/mgrs/mgrs_full.fgb’ saved [13787464/13787464]\n",
+      "\n"
+     ]
+    }
+   ],
+   "source": [
+    "import geopandas as gpd\n",
+    "\n",
+    "! wget https://clay-mgrs-samples.s3.amazonaws.com/mgrs_full.fgb -O data/mgrs/mgrs_full.fgb"
```
weiji14, line 62: Should we mention the original KML file this `mgrs_full.fgb` file was converted from? Assuming that you got it from https://sentinel.esa.int/web/sentinel/missions/sentinel-2/data-products at https://sentinel.esa.int/documents/247904/1955685/S2A_OPER_GIP_TILPAR_MPC__20151209T095117_V20150622T000000_21000101T000000_B00.kml?

docs/clay_over_aoi.ipynb
```diff
@@ -0,0 +1,727 @@
...
+      "\n",
+      "KeyboardInterrupt\n",
+      "\n"
+     ]
+    }
+   ],
+   "source": [
+    "for index, row in mgrs_aoi.iterrows():\n",
+    "    print(index, row)\n",
+    "    ! python scripts/datacube.py --sample data/mgrs/mgrs_aoi.fgb --subset 1500,1500,2524,2524 --localpath data/chips  --index {index} --dateranges 2020-01-01/2020-04-01,2021-06-01/2021-09-15"
```
weiji14, line 595: This line requires a `PC_SDK_SUBSCRIPTION_KEY` token to be set I think, otherwise users will get a `requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://planetarycomputer.microsoft.com/api/sas/v1/token/sentinel1euwestrtc/sentinel1-grd-rtc`. Need to document this inline perhaps, xref #119.

weiji14 (approved):
Thanks @yellowcap, just some minor recommendations/questions below. I think we're trying to move fast, so I'll pre-approve this, but feel free to handle any of the suggestions below if you have time before merging.

docs/clay_over_aoi.ipynb
```diff
@@ -0,0 +1,727 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "1f640c20-87b3-445b-9de8-e91670562df1",
+   "metadata": {},
+   "source": [
+    "## Create the embeddings for each training chip\n",
+    "\n",
+    "The checkpoints can be downloaded from huggingface."
```
weiji14, line 605: Link to HuggingFace page perhaps?

```suggestion
    "The checkpoints can be downloaded from huggingface at https://huggingface.co/made-with-clay/Clay."
```

docs/clay_over_aoi.ipynb
weiji14, line 690: Lightning supports passing in a checkpoint URL (see https://lightning.ai/docs/pytorch/2.1.2/common/checkpointing_advanced.html#resume-training-from-a-cloud-checkpoint), so can remove the `wget` line above and predict directly from the HuggingFace checkpoint URL. The checkpoint will be downloaded automatically to `~/.cache/torch/hub/checkpoints/`

```suggestion
    "! python trainer.py predict --ckpt_path=https://huggingface.co/made-with-clay/Clay/resolve/main/Clay_v0.1_epoch-24_val-loss-0.46.ckpt \\\n",
```

yellowcap: Most definitively, we can remove this line once the trainer has been udpated.

scripts/datacube.py
yellowcap: Good catch, I will update accordingly.

scripts/datacube.py
yellowcap: I see this is creating confusion, I'll try to work this better in the help string.

The goal of this is to reduce the amount of data when processing in test mode. Normally, the script would download the entire scene, which is 10k pixels squared. Whit this one can reduce the amount of pixels downloaded and passed to the tiler script. So this is cutting a window from the scene data. The tiler will then work on this smaller input, but be unaltered otherwise.

brunosan: Not to make it more complex, but I understand we download the entire MGRS tile, even when many of the  chips might be outside of the requested AOI. If we drop chips outside of the AoI, we can just use a point or small polygon for testing.

docs/clay_over_aoi.ipynb
yellowcap: Good idea, the full sample file is constructed in the [landcover.sh](https://github.com/Clay-foundation/model/blob/main/scripts/landcover.sh) script. There I actually got it from the NASA site but its the same file. I'll link to this in the notebook.

docs/clay_over_aoi.ipynb
yellowcap: Added reference to the markdown

docs/clay_over_aoi.ipynb
yellowcap: Good idea, added.

scripts/tile.py
yellowcap: Separate PR, coming up.

scripts/datacube.py
yellowcap: That is correct and would be the best approach, but for that the datacube.py script needs much more refactoring. Currently it's fully focused on downloading the entire MGRS tile. Filtering that spatially will require passing the AOI itself and handling spatial intersections, so that indeed is quite a bit more complicated.

For now I can state also that more clearly, that the AOI approach right now means getting data for full MGRS tiles that intersect with the AOI.

docs/clay_over_aoi.ipynb
yellowcap: Good point, changed to use `shapely.box`

PR #122, opened 2024-01-15
Convert run-over-aoi ipynb to md

yellowcap:
The ipynb will not run in build because its too long to run. So this needs to be frozen into md for the docs.

This also applies to the other ipynb in the docs most likely.


docs/run_region.md
```diff
@@ -0,0 +1,124 @@
...
+## Prepare folder strucutre for data
+
+```bash
+# Move into the model repository
+cd /path/to/repository/model/
+
+# Ensure data sub-directories exist
+mdkir data/mgrs
+mdkir data/chips
+mdkir data/embeddings
```
weiji14, line 15:
```suggestion
mkdir data/mgrs
mkdir data/chips
mkdir data/embeddings
```

weiji14 (approved):
Thanks @yellowcap, much easier to review the markdown! Just some typos on the `mkdir` command, otherwise ok.

Btw, we can also set `execute_notebooks: 'cache'` here:
https://github.com/Clay-foundation/model/blob/a4377ccd674d5770a000db53bc7334805995e8f4/docs/_config.yml#L8C1-L11C27

This will only run the notebooks if there are no outputs in the cells. See https://jupyterbook.org/en/stable/content/execute.html for more details, there are other options such as excluding specific files from being run.

PR #129, opened 2024-01-18
Add notebooks to show internal working on CLAY models

srmsoumya:
Add notebooks to show how to:
- reconstruct the pixel space
- visualize location embeddings
- visualize the embeddings from the model
- interpolate in the embedding space

Add `shuffle` as an argument to clay, useful while inferencing when we don't want to shuffle the input to the model.


docs/animate/sample.gif
srmsoumya, line 1: @weiji14 , JupyterBook is executing the notebook every time we make a change, which results in a new version of the `sample.gif` file being saved. 
This is increasing the repository size unnecessarily. How can we execute the notebooks only when necessary to avoid this?

weiji14: Aren't we gitignoring *.git files here?

https://github.com/Clay-foundation/model/blob/ae70345395bb541403fda295dadb04fd2b3e191d/.gitignore#L21

There's a way to prevent execution of certain notebooks following https://jupyterbook.org/en/stable/content/execute.html#exclude-files-from-execution. We could provide a list of notebooks that we know would take a lot of time or resources to run.

yellowcap:
Closing this as out of date. Feel free to reopen @srmsoumya if you think we should keep this or parts of this.

Issue #130, opened 2024-01-19
Update `train.py` script to optionally store the raw encoder output


yellowcap:
To test how raw encodings look like, and to potentially store raw encoder output during batch inference for speeding up downstream applications.

Refs https://github.com/Clay-foundation/model/issues/127

brunosan:
Also to potentially increase the semantic spatial resolution 4x to 16x. #110

Issue #123, opened 2024-01-16
Disable patch shuffling by default


yellowcap:
Shuffling the patches is important during MAE training, but is confusing and requires extra attention when fine tuning or during inference.

We should have an automatic way of handling this, so that the downstream user does not have to worry about the shuffling at all.

Refs

https://github.com/Clay-foundation/model/blob/2a7a6e2b63ba453ab73a192f8560bd724d4f98d5/src/model_clay.py#L598

https://github.com/Clay-foundation/model/pull/118#discussion_r1451790971

yellowcap:
Could we make this automatic and dependent on inference vs training? Seems like during training it should be always `True`, and during inference there is no downside of having it always `False`. Or am I not seeing something here?

Issue #127, opened 2024-01-18
Run v0.1 embeddings to store raw encoder output


yellowcap:
To enable fast downstream applications, we could store the raw encoder output and not only the average embedding that we are already creating for similarity search.

Refs

- https://github.com/Clay-foundation/model/issues/107

brunosan:
Thanks.
From the discussion, I also realize that we don't really have 1 embedding per self-attention patch, we have 1 per self attention AND layer group. If that's the case, we should add it somewhere on the documentation.

I.e. 
1. Our big MGRS tile is split into `512x512` patches. (at Sentinel resolution means `~5kmx5km`
2. Each patch is further split into a "self-attention patch" of size [`32x32`](https://github.com/Clay-foundation/model/blob/0e34498fbde6b4671250b2fda649aec56a3ea0d3/src/model_clay.py#L799) or `~300mx300m`.
3. For each self-attention patch, the 13 layers are grouped into [6 groups](https://github.com/Clay-foundation/model/blob/main/src/model_clay.py#L584), and we create one embedding per group.

Note: These grouping do NOT mean that the model has parallel track for each group. When training we calculate the self-attention (QKV) individually for each layer. Groups, are more akin to sentences, groups of words. This means that the RGB group also has information about that SAR has, and vice versa. 

@srmsoumya to confirm this. If this is the case, I don't understand the value of the grouping instead of making one embedding per self-attention patch. What's the value of grouping the embeddings this way?

TLDR for @MaceGrim.  The semantic resolution at the self-attention path before the average is `~300m` but also split into dominant groups of bands.

brunosan:
Thanks.
From the discussion, I also realize that we don't really have 1 embedding per self-attention patch, we have 1 per self attention AND layer group. If that's the case, we should add it somewhere on the documentation.

I.e. 
1. Our big MGRS tile is split into `512x512` patches. (at Sentinel resolution means `~5kmx5km`
2. Each patch is further split into a "self-attention patch" of size [`32x32`](https://github.com/Clay-foundation/model/blob/0e34498fbde6b4671250b2fda649aec56a3ea0d3/src/model_clay.py#L799) or `~300mx300m`.
3. For each self-attention patch, the 13 layers are grouped into [6 groups](https://github.com/Clay-foundation/model/blob/main/src/model_clay.py#L584), and we create one embedding per group.

[...]

PR #133, opened 2024-01-23
Add option to output raw patch embeddings

yellowcap:
This PR adds an option to the Lightning CLI to output patch level embeddings. I.e. one embedding per patch in each chip. The band group dimension is reduced and, so the patch embeddings are averages over the band groups.

This adds two args to the CLI `output_patch_embeddings`, and `shuffle`. Because for the patch embeddings we need to ensure that shuffle is off, while it is on by default. See also https://github.com/Clay-foundation/model/issues/123.

Updated the documentation to explain these changes.

This PR Closes https://github.com/Clay-foundation/model/issues/130

### DIFF
--- a/docs/model_embeddings.md
+++ b/docs/model_embeddings.md
@@ -45,8 +45,7 @@ Step by step instructions to create embeddings for a single MGRS tile location
                              --data.data_dir=s3://clay-tiles-02/02/27WXN \
                              --data.batch_size=32 \
                              --data.num_workers=16 \
-                             --model.embeddings_level=group \
-                             --model.shuffle=False
+                             --model.embeddings_level=group
    ```
 
    This should output a GeoParquet file containing the embeddings for MGRS tile
@@ -65,9 +64,6 @@ Step by step instructions to create embeddings for a single MGRS tile location
    The embeddings are flattened into one dimensional arrays because pandas
    does not allow for multidimensional arrays. This makes it necessary to
    reshape the flattened arrays to access the patch level embeddings.
-
-   The `shuffle` flag should be set to `False` of the `output_patch_embeddings` is
-   `True` to ensure that the order of the patch embeddings is not randomized.
 
    ```{note}
    For those interested in how the embeddings were computed, the predict step
--- a/docs/model_embeddings.md
+++ b/docs/model_embeddings.md
@@ -45,7 +45,7 @@ Step by step instructions to create embeddings for a single MGRS tile location
                              --data.data_dir=s3://clay-tiles-02/02/27WXN \
                              --data.batch_size=32 \
                              --data.num_workers=16 \
-                             --model.output_patch_embeddings=False \
+                             --model.embeddings_level=group \
                              --model.shuffle=False
    ```
 
@@ -54,10 +54,17 @@ Step by step instructions to create embeddings for a single MGRS tile location
    smaller 512x512 chips), saved to the `data/embeddings/` folder. See the next
    sub-section for details about the embeddings file.
 
-   The `output_patch_embeddings` flag determines how the embeddings are calculated.
-   If `False`, one average embedding per MGRS tile of size 768 will be created. If
-   `True`, the embeddings will be kept at the patch level. The embedding array will
-   be of size 16 * 16 * 768, representing.
+   The `embeddings_level` flag determines how the embeddings are calculated.
+   The default is `mean`, resulting in one average embedding per MGRS tile of
+   size 768. If set to `patch`, the embeddings will be kept at the patch level.
+   The embedding array will be of size 16 * 16 * 768, representing one
+   embedding per patch. The third option `group` will keep the full
+   dimensionality of the encoder output, including the band group
+   dimension. The array size of those embeddings is 6 * 16 * 16 * 768.
+
+   The embeddings are flattened into one dimensional arrays because pandas
+   does not allow for multidimensional arrays. This makes it necessary to
+   reshape the flattened arrays to access the patch level embeddings.
 
    The `shuffle` flag should be set to `False` of the `output_patch_embeddings` is
    `True` to ensure that the order of the patch embeddings is not randomized.
@@ -115,7 +122,8 @@ and contains a record of the embeddings, spatiotemporal metadata, and a link to
 the GeoTIFF file used as the source image for the embedding. The table looks
 something like this:
 
-Embedding size is 768 by default, and 16 * 16 * 768 for patch level embeddings.
+Embedding size is 768 by default, 16 * 16 * 768 for patch level embeddings, and
+6 * 16 * 16 * 768 for group level embeddings.
 
 
 |         source_url          |    date    |    embeddings    |   geometry   |
@@ -167,6 +175,7 @@ To convert each row into patch level embeddings, the embedding array has to
 be unravelled into 256 patches like so
 
 ```{code}
+# This assumes embeddings levels set to "patch"
 ravelled_patch_embeddings = geodataframe.embeddings[0]
-patch_embeddings = ravelled_patch_embeddings.reshape(256, 768)
+patch_embeddings = ravelled_patch_embeddings.reshape(16, 16, 768)
 ```
--- a/docs/model_embeddings.md
+++ b/docs/model_embeddings.md
@@ -44,8 +44,8 @@ Step by step instructions to create embeddings for a single MGRS tile location
                              --trainer.precision=bf16-mixed \
                              --data.data_dir=s3://clay-tiles-02/02/27WXN \
                              --data.batch_size=32 \
-                             --data.num_workers=16
-                             --output-patch-embeddings=False
+                             --data.num_workers=16 \
+                             --output-patch-embeddings=False \
                              --shuffle=False
    ```
 
@@ -155,4 +155,18 @@ print(geodataframe)
 Further reading:
 - https://guide.cloudnativegeo.org/geoparquet
 - https://cloudnativegeo.org/blog/2023/10/the-geoparquet-ecosystem-at-1.0.0
+```
+
+## Converting to patch level embeddings
+
+In the case where patch level embeddings are requested, the resulting array
+will have all patch embeddings ravelled in one row. Each row represents a
+512x512 pixel image, and contains 16x16 patch embeddings.
+
+To convert each row into patch level embeddings, the embedding array has to
+be split
+
+```{code}
+ravelled_patch_embeddings = geodataframe.embeddings[0]
+patch_embeddings = ravelled_patch_embeddings.reshape(256, 768)
 ```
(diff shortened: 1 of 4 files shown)


src/model_clay.py
```diff
@@ -892,13 +893,25 @@ def predict_step(
...
-        # Take the mean of the embeddings along the sequence_length dimension
-        # excluding the last two latlon_ and time_ embeddings, i.e. compute
-        # mean over patch embeddings only
-        embeddings_mean: torch.Tensor = embeddings_raw[:, :-2, :].mean(dim=1)
-        assert embeddings_mean.shape == torch.Size(
-            [self.model.encoder.B, 768]  # (batch_size, hidden_size)
-        )
+        if self.hparams.output_patch_embeddings:
+            embeddings_raw = rearrange(
+                embeddings_raw[:, :-2, :], "b (w h g) s -> b (w h s) g", w=16, h=16, g=6
```
srmsoumya, line 898: The order of the tensor matters, in this case the ouput from the encoder i.e embeddings_raw is of shape `batch x (group x  num_spatial_patches) x embedding_dims`.

So, the einops operation should be, `b (g l) d -> b g l d` - you can check this notebook for reference: https://github.com/Clay-foundation/model/blob/docs/model/docs/clay-v0-visualization.ipynb

src/model_clay.py
```diff
@@ -892,13 +893,25 @@ def predict_step(
...
+            )
+            embeddings_mean: torch.Tensor = embeddings_raw.mean(dim=2)
```
srmsoumya, line 900: We can use the `reduce` operation from einops, that is more explicit.

For eg: `reduce(embeddings_raw, "b g h w d -> b h w d", "mean")`

srmsoumya (changes requested):

src/model_clay.py
yellowcap: good idea, Einstein is the solution for everything 🧑‍🔬  😸

src/model_clay.py
yellowcap: Done, but please review again to make sure I got it right this time. Related to this, do you think [this](https://github.com/Clay-foundation/model/pull/133/files#diff-2f39d0012d2540cdab1ac1ff882a3875d829b85f0acee68fe102e6b8a4f2c6c0R171) is a good way to unravel the patch embeddings?

docs/model_embeddings.md
```diff
@@ -44,14 +44,24 @@ Step by step instructions to create embeddings for a single MGRS tile location
                              --trainer.precision=bf16-mixed \
                              --data.data_dir=s3://clay-tiles-02/02/27WXN \
                              --data.batch_size=32 \
-                             --data.num_workers=16
+                             --data.num_workers=16 \
+                             --output-patch-embeddings=False \
+                             --shuffle=False
```
weiji14, line 49: Does this work? I thought we have to pass `--model.init_args.shuffle=False` (untested).

```suggestion
                             --model.init_args.output_patch_embeddings=False \
                             --model.init_args.=False
```

src/model_clay.py
weiji14: Mentioned this below at https://github.com/Clay-foundation/model/pull/133#discussion_r1464236962, but can we just keep the unravelled shape (B, 256, 768), or (B, 1536, 768)? Less work for the downstream user since they won't need to figure out how to unravel the tensor.

src/model_clay.py
```diff
@@ -892,13 +895,28 @@ def predict_step(
...
+            embeddings_raw = rearrange(
+                embeddings_raw[:, :-2, :], "b (g l) s -> b g (l s)", l=256, g=6
+            )
+            embeddings_mean = reduce(embeddings_raw, "b g s -> b s", "mean")
+            assert embeddings_mean.shape == torch.Size(
+                [
+                    self.model.encoder.B,
+                    256 * 768,
+                ]  # (batch_size, nr of patches * hidden_size)
+            )
```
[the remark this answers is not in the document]
yellowcap: Unsqueezed would undoubtedly be better, but I assumed that we can only store one-dimensional arrays in a field. I guess from your questions that is not the case. Can we store multidimensional arrays here? If yes, lets store the unsqueezed one for sure!

src/model_clay.py
yellowcap: Yes absolutely. Don't know why I thought it had to be a 1d array...

docs/model_embeddings.md
yellowcap: Looks like the correct argument is `model.output_patch_embeddings`

```
>>> python trainer.py fit -h | grep embeddings
...                                [--model.output_patch_embeddings OUTPUT_PATCH_EMBEDDINGS]
...  --model.output_patch_embeddings OUTPUT_PATCH_EMBEDDINGS
```

src/model_clay.py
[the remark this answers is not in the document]
weiji14: Yes, GeoParquet/Arrow technically supports [FixedShapeTensorArray](https://arrow.apache.org/docs/14.0/python/generated/pyarrow.FixedShapeTensorArray.html) which can be multi-dimensional.

src/model_clay.py
weiji14: That said, do we want to store a (256, 768) embedding tensor in a single row, or split it into 256 rows of 768-length embeddings? I'm reading the original thread at #127, and it sounds like we need to enable searching on the 32x32 patch level instead of 256x256 chips, which might mean storing a row for each patch?

We can store 2D arrays in GeoParquet, but I'm not sure if vector databases allow indexing 2D arrays, or if we need to make it 1D. Note that this would increase the embedding file size significantly (x256), and the vector database indexing will be much slower. But if these raw embeddings are only meant to be generated on an ad-hoc basis for small locations, and not for similarity search applications, it should be fine.

yellowcap: I thought about that too, but that would also imply that we have to track the MGRS tile and the source path 256x more. That seems ineficient. My thinking was that the splitting into rows can happen at the moment on ingestion to a vector search DB or when analysing if required. For "local manual inspection" situations, having the patches in a single array is useful too. So I vote to keep it at one row in the gpd and let the separation be a downstream issue.

src/model_clay.py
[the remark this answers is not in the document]
yellowcap: That will require a rewrite without pandas, because pandas can not handle this and implicitly flattens the `FixedShapeTensorArray`, see example below. I ran across this when running tests and not getting the expected 2D arrays. This defintively complicates things in terms of keeping the structure.

```
In [15]: array = np.arange(8).reshape((2,2,2))
    ...: 
    ...: arrow = pa.FixedShapeTensorArray.from_numpy_ndarray(array)
    ...: 
    ...: df = pd.DataFrame(data={"arrow": arrow})
    ...: 
    ...: df
    ...: 
Out[15]: 
          arrow
0  [0, 1, 2, 3]
1  [4, 5, 6, 7]
```

docs/model_embeddings.md
```diff
@@ -44,14 +44,24 @@ Step by step instructions to create embeddings for a single MGRS tile location
...
 
    This should output a GeoParquet file containing the embeddings for MGRS tile
    27WXN (recall that each 10000x10000 pixel MGRS tile contains hundreds of
    smaller 512x512 chips), saved to the `data/embeddings/` folder. See the next
    sub-section for details about the embeddings file.
 
+   The `output_patch_embeddings` flag determines how the embeddings are calculated.
+   If `False`, one average embedding per MGRS tile of size 768 will be created. If
+   `True`, the embeddings will be kept at the patch level. The embedding array will
+   be of size 16 * 16 * 768, representing.
```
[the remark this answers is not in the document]
yellowcap: Updated wording to adopt to the embedding levels input.

src/model_clay.py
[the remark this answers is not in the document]
yellowcap: Ok so based on the above, the arrays are outputted in a flat structure for now. I updated the documentation with a sentence about where the structure comes from and why the arrays are flat.

https://github.com/Clay-foundation/model/pull/133/files#diff-2f39d0012d2540cdab1ac1ff882a3875d829b85f0acee68fe102e6b8a4f2c6c0R60

This is not ideal but keeping the multidimensional arrays now would require refactoring of how we construct the geoparquet files.

src/tests/test_model.py
```diff
@@ -87,19 +87,28 @@ def test_model_vit_fit(datapipe):
 @pytest.mark.parametrize(
     "litmodule,precision",
     [
-        (CLAYModule, "bf16-mixed" if torch.cuda.is_available() else "32-true"),
-        (ViTLitModule, "bf16-mixed"),
+        (CLAYModule, "16-mixed" if torch.cuda.is_available() else "32-true"),
```
yellowcap, line 90: Is this ok? My GPU does not support `bf16-mixed`, so if it does not matter this makes my life easier when testing locally.

docs/model_embeddings.md
```diff
@@ -44,14 +44,24 @@ Step by step instructions to create embeddings for a single MGRS tile location
...
+
+   The `shuffle` flag should be set to `False` of the `output_patch_embeddings` is
+   `True` to ensure that the order of the patch embeddings is not randomized.
```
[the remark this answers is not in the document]
yellowcap: Proposed solution here https://github.com/Clay-foundation/model/pull/135

src/model_clay.py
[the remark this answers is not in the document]
weiji14: Hmm, have you tried inserting a FixedShapeTensorArray into a row cell directly? This seems to work:

```python
import numpy as np
import pandas as pd
import pyarrow as pa

array0 = np.arange(0, 8).reshape((2, 2, 2))
array1 = np.arange(9, 17).reshape((2, 2, 2))

arrow0 = pa.FixedShapeTensorArray.from_numpy_ndarray(array0)
arrow1 = pa.FixedShapeTensorArray.from_numpy_ndarray(array1)

df = pd.DataFrame(data={"embeddings": [arrow0, arrow1]})
print(df)
#                             embeddings
# 0         ([0, 1, 2, 3], [4, 5, 6, 7])
# 1  ([9, 10, 11, 12], [13, 14, 15, 16])
```

src/tests/test_model.py
weiji14: Oh yes, should be ok to change this since it's only on the unit test. But we'll still use `bf16-mixed` when training the actual Foundation Model on A10G GPUs.

src/model_clay.py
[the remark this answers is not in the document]
yellowcap: The arrow part still flattens the arrays that are input. For this to work, we would have to do the manual deconstruction up to 3 dimensions deep. Not sure that this will make things more user friendly. With this approach, the user would have to reconstruct a list of a list of a list of arrays in the "group" case.

src/tests/test_model.py
```diff
@@ -87,19 +87,27 @@ def test_model_vit_fit(datapipe):
...
```
srmsoumya, line 90: Is `bf16-mixed` precision giving any issue while inferencing?

srmsoumya (approved):
Like the idea of storing embeddings at multiple levels.

src/tests/test_model.py
yellowcap: It is for the Nvidia Geforce RTX on my local machine, which does not support this type apparently.

PR #135, opened 2024-01-25
Shuffle patches only during training

### LINKED ISSUE #123: Disable patch shuffling by default

yellowcap:
This removes the shuffle argument and shuffles patches only during training. Closes #123

### DIFF
--- a/src/model_clay.py
+++ b/src/model_clay.py
@@ -56,7 +56,6 @@ class Encoder(nn.Module):
         mask_ratio,
         image_size,
         patch_size,
-        shuffle,
         dim,
         depth,
         heads,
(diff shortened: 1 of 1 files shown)


srmsoumya (approved):
Good idea to not shuffle the patches by default.

PR #139, opened 2024-01-26
Initial version of minicube

yellowcap:
This script should probably be divided into functions and converted to cli tool. But it should work more or less as is.

### DIFF
--- a/scripts/minicube.py
+++ b/scripts/minicube.py
@@ -1,14 +1,24 @@
 from pathlib import Path
 
 import geopandas as gpd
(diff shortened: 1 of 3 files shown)


yellowcap:
This can be combined with the partial input script that is also working but not yet final

https://github.com/Clay-foundation/model/blob/docs/model/docs/clay-v0-partial-inputs.ipynb

yellowcap:
Close in favor of https://github.com/Clay-foundation/model/pull/149 which integrates the minicube script in a notebook.

PR #149, opened 2024-02-05
Tutorial on burn scar analysis using embeddings from partial inputs

yellowcap:
This PR adds a tutorial on Burn scar analysis using embeddings from partial inputs

This notebook contains a complete example for how to run Clay. It combines the following three different aspects

1. Create single-chip datacubes with time series data for a location and a date range
2. Run the model with partial inputs, in this case RGB + NIR
3. Study burn scares through the embeddings generated for that datacube

Closes #126
Closes #136

PR #154, opened 2024-02-09
Partial inputs - Pakistan flood tutorial

lillythomas:
This is a tutorial following the style of #149 to show how to generate embeddings from partial inputs and find a signal for a [major monsoon flood in Padidan, Pakistan (August 2022)](https://floodlist.com/asia/pakistan-monsoon-floods-august-2022) with PCA and t-SNE.

Modifications (aside from AOI and time):
- Use of SWIR band group
- Use of the correct mean and std for the nir, not nir08 band
- Addition of t-SNE computation
- modification to the datamodule that enables the non-normalized dates to be added to the batch dictionary (for plotting with the embeddings when there are 2 or more components involved...aka, date can't be one of the axes)


src/datamodule.py
```diff
@@ -60,7 +61,14 @@ def read_chip(self, chip_path):
...
         date = chip.tags()["date"]  # YYYY-MM-DD
-        year, month, day = self.normalize_timestamp(date)
+        (
+            year,
+            month,
+            day,
+            year_non_norm,
+            month_non_norm,
+            day_non_norm,
+        ) = self.normalize_timestamp(date)
```
weiji14, line 71: We are already returning the non-normalized date as a YYYY-MM-DD string no? Instead of using `ts.append(batch["timestep_non_norm"])` in in your partial-inputs-flood-tutorial.ipynb notebook, could you use `ts.append(batch["date"])`?

src/datamodule.py
```diff
@@ -175,7 +184,7 @@ def train_dataloader(self):
             self.trn_ds,
             batch_size=self.batch_size,
             num_workers=self.num_workers,
-            shuffle=True,
+            shuffle=False,
```
weiji14, line 187: We should be shuffling the mini-batches in the `train_dataloader`, is there a reason to set it to False here?
```suggestion
            shuffle=True,
```

src/datamodule.py
lillythomas: Fixed this. It worked, thanks!

src/datamodule.py
lillythomas: I rolled back changes to the datamodule in this PR, so it is the same as in main. I think this should be examined in a separate PR since no training is occurring for the purpose of this one.

weiji14 (approved):

Issue #161, opened 2024-02-23
EPIC: Support running on macOS


weiji14:
To allow Clay to run on not just Linux operating systems, but OSX as well!

This comes from user feedback during DevSeed's Team Week, where we found out that the folks developing the Clay model are predominantly Linux-based, but users were using other operating systems like macOS.

Non-OS specific:
- [ ] Document how to install mamba (using brew on OSX, or add non-OS specific instructions)
- [ ] Extra steps to handle mamba commands on zsh
- [ ] Create a docker image for Clay (@leothomas) #166

OSX (Intel chips):
- [x] Solve installation on OSX by removing cuda part which is Linux specific (@chuckwondo)
- [x] Get CI running on OSX 10.14+
- [ ] Mention minimum version of pytorch needed is 2.1?

OSX-ARM64 (M1/M2/M3 chips):
-  [x] Get CI running on `macos-14` (#164) ~Figure out why `torchdata~=0.7.1` and `torchvision~=0.16.1` cannot be installed on `osx-arm64`, maybe https://github.com/conda-forge/torchvision-feedstock/pull/89 will help?~

mikeskaug:
I would be happy to help with this if there is anything I can do. A few weeks ago I started working on getting things set up and running on my Mac. I was interested in experimenting with the project and my typical approach is to start local. My first goal was to get to the point where I could train and predict on a small data set and then move on from there. I haven't finished but I would be happy to share my approach or issues I've encountered.

I've made a few minor changes but the most significant departure has been to use pip instead of mamba to install requirements. I don't know if that's useful or just a bad idea, but I thought it would be worth a try and it seems to work. (Although I guess I'm not 100% sure until I have the various cli commands running, which currently seem to be failing because of data type support on cpu.)

weiji14:
Hi @mikeskaug, thanks for offering to help! If you can post some of the error messages you're getting (open a bug report issue), that would be great for us to keep track. There's been lots of dtype related bugs (e.g. https://github.com/Clay-foundation/model/issues/134) due to different operating systems and GPU/CPU devices being used, and we'd like to iron those out as much as possible!

If you're keen to contribute, we're open to help with improving our documentation, or you can also jump in and review some of our outstanding PRs (e.g. #164 and #166 are macOS related ones). Or if you have a use-case/dataset you'd like to share and collaborate on, we can discuss more too :smile:

ritwikvashistha:
[...]

PR #162, opened 2024-02-23
Add osx platform to conda lock file and CI

chuckwondo:
Addresses https://github.com/Clay-foundation/model/issues/161 regarding OSX/CUDA incompatability. Also adds macos-12 to os matrix in GitHub CI workflow.


weiji14 (approved):
Thank you!

PR #163, opened 2024-02-23
Move pipeline scripts into dedicated folder

yellowcap:
This is to prepare for adding more scripts to the scripts folder that are not pipeline related.


srmsoumya (approved):
Looks all good.

PR #164, opened 2024-02-26
Add osx-arm64 platform to conda-lock.yml file and GitHub Actions CI

weiji14:
Support installation on macOS ARM64 devices (M1 chips) by:

- Adding `osx-arm64` as another platform in the `environment.yml` file
- Relocking the conda-lock.yml file using `conda-lock lock --mamba --file environment.yml --with-cuda=12.0`
- Adding `macos-14` to the `.github/workflows/test.yml` GitHub Actions workflow

References:
- https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source

Addresses https://github.com/Clay-foundation/model/issues/161 and extends #162.

PR #166, opened 2024-02-26
Add a docker container (and docker-compose file) to run the model/notebooks in a containerize environment

leothomas:
In order to make it easier to run the model/notebooks without having to manage installing dependencies across various machines/environments, I've added a `micromamba` based Dockerfile which will create the conda environment with the specified libraries. 

I've also added a docker compose file, in order to specify the build-time and run-time arguments for exposing the jupyter lab port and mounting the current directly as a volume, into the docker container. This will allow users to modify any of the model/notebook code locally, without having to re-build the image. 

By default the docker image starts with running jupyter lab but this can be overridden both in the docker-compose or even in the command line with any other python or bash command. 

The `platform=linux/amd64` build and run-time args enable the image to be built on Mac M1 while maintaining compatibility with Linux. 

The container can be run with: 
`docker-compose up` or `docker-compose run claymodel <command>` where `command` is a command which override the jupyter lab startup

The container can also be built directly (bypassing the need for docker-compose) with: 
```bash
docker build . -t clay --platform linux/amd64
```
and then run with: 
```bash
docker run --rm -it -v $(pwd):/model -p 8888:8888 -e ENV_NAME=claymodel --platform linux/amd64 clay:latest
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,26 @@
+# FROM condaforge/miniforge3:latest
+
+# WORKDIR /model
+
+# COPY . .
+
+# RUN mamba install -y -n base --file environment.yml
+
+# #RUN mamba activate claymodel
+# #RUN echo "mamba activate ${ENV_NAME}" >> ~/.bash_profile
+
+# ENTRYPOINT [ "bash", "-l", "-c" ]
+
+# CMD ["jupyter-lab", "--ip", "0.0.0.0", "--port", "8888", "--allow-root"]
+
+FROM mambaorg/micromamba:1.5.6 
+
+WORKDIR /model
+
+COPY --chown=$MAMBA_USER:$MAMBA_USER . .
+
+RUN micromamba create -y -n claymodel --file environment.yml && \
+    micromamba clean --all --yes
+
+ENTRYPOINT ["/usr/local/bin/_entrypoint.sh"]
+CMD ["jupyter-lab", "--ip", "0.0.0.0", "--port", "8888", "--allow-root"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,16 @@
+version: '3.8'
+
+services:
+  claymodel:
+    build: 
+      context: .
+      dockerfile: Dockerfile
+      args: 
+        - platform=linux/amd64 # Specify the target platform for build
+    platform: linux/amd64 # Specify the target platform for run
+    ports:
+      - "8888:8888" # Expose port 8888 on the host and bind it to port 8888 in the container
+    volumes:
+      - .:/model # Mount the current directory to /app in the container
+    environment:
+      - ENV_NAME=claymodel # Ensure the claymodel conda environment is activated
--- a/environment.yml
+++ b/environment.yml
@@ -14,7 +14,8 @@ dependencies:
   - lightning~=2.1.0
   - matplotlib-base~=3.8.2
   - planetary-computer~=1.0.0
-  - pytorch~=2.1.0=*cuda120*
+  - pytorch~=2.1.0
+  - pyarrow~=15.0.0
   - python=3.11
   - rioxarray~=0.15.0
   - scikit-image~=0.22.0
@@ -26,3 +27,4 @@ dependencies:
   - vit-pytorch~=1.6.4
   - wandb~=0.15.12
   - zarr~=2.16.1
+  


environment.yml
```diff
@@ -14,7 +14,8 @@ dependencies:
   - lightning~=2.1.0
   - matplotlib-base~=3.8.2
   - planetary-computer~=1.0.0
-  - pytorch~=2.1.0=*cuda120*
+  - pytorch~=2.1.0
+  - pyarrow~=15.0.0
```
weiji14, line 18: Could you resolve the merge conflict here with the `main` branch? Also see if removing the `pyarrow` pin works. I managed to get conda to solve for `osx-arm64` recently on https://github.com/Clay-foundation/model/pull/164 after https://github.com/conda-forge/torchvision-feedstock/pull/89/files was merged.

Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+WORKDIR /model
+
+COPY --chown=$MAMBA_USER:$MAMBA_USER . .
+
+RUN micromamba create -y -n claymodel --file environment.yml && \
+    micromamba clean --all --yes
+
+ENTRYPOINT ["/usr/local/bin/_entrypoint.sh"]
+CMD ["jupyter-lab", "--ip", "0.0.0.0", "--port", "8888", "--allow-root"]
```
weiji14, line 11: Is it possible to avoid the `--allow-root` somehow?

docker-compose.yml
```diff
@@ -0,0 +1,16 @@
...
+    build: 
+      context: .
+      dockerfile: Dockerfile
+      args: 
+        - platform=linux/amd64 # Specify the target platform for build
+    platform: linux/amd64 # Specify the target platform for run 
+    ports:
+      - "8888:8888" # Expose port 8888 on the host and bind it to port 8888 in the container
+    volumes:
+      - .:/model # Mount the current directory to /app in the container
```
weiji14, line 14:
```suggestion
      - .:/model # Mount the current directory to /model in the container
```

weiji14 (commented):
Thanks @leothomas! Just some small comments for now. Do you think we should also add a `.dockerignore` file to keep the Docker image a pure virtual environment? Also, what are your thoughts on setting up some CI to push pre-built containers to a docker registry (can be done in a separate PR)?

chuckwondo (changes requested):
In addition to the individual comments/suggestions, please also add a section to `README.md` about how to run the Docker container (`docker compose up`) as an _alternative_ to installing things locally, and how to access JupyterLab once the container is started. In particular, mention the URLs that appears in the logging messages:
```
claymodel-1  |     Or copy and paste one of these URLs:
claymodel-1  |         http://ffd23ea64b9b:8888/lab?token=abebb7b9476b8fff1fb8b543cc552e8c9641b5b38547c3cb
claymodel-1  |         http://127.0.0.1:8888/lab?token=abebb7b9476b8fff1fb8b543cc552e8c9641b5b38547c3cb
```

environment.yml
chuckwondo: Agreed. You should be able to simply ditch your changes to `environment.yml`. If not, then keep in mind that whenever you modify `environment.yml`, you must regenerate `conda-lock.yml`, so you should never be committing changes to _only_ `environment.yml`.

chuckwondo:
> > Thanks @leothomas! Just some small comments for now. Do you think we should also add a `.dockerignore` file to keep the Docker image a pure virtual environment? Also, what are your thoughts on setting up some CI to push pre-built containers to a docker registry (can be done in a separate PR)?
> 
> Absolutely add a `.dockerignore` file to this PR. You can probably start with much (or all) of what's in `.gitignore`.
> 
> However, keep in mind that there are cases where you must be more explicit in `.dockerignore`.
> 
> For example, you cannot use only `__pycache__/` in `.dockerignore` because that will ignore _only_ a top level `__pycache__/` directory. To ignore such a directory at _all_ levels, you must use `**/__pycache__/` in `.dockerignore`. Another good candidate for this is to add `**/.ipynb_checkpoints/` since there are notebooks in the `docs` directory.

Alternatively, as I mentioned to you ages ago, I tend to "invert" my use of `.dockerignore` to make it act more like an "allow" list rather than a "deny" list, which I find to be safer and clearer. For example, consider making a `.dockerignore` that ignores _everything_ and then does _not_ ignore the things you want to "allow":

```
*
!environment.yml
!conda-lock.yml
!**/*.py
!**/*.ipynb
!**/*.sh
```

environment.yml
```diff
@@ -26,3 +27,4 @@ dependencies:
   - vit-pytorch~=1.6.4
   - wandb~=0.15.12
   - zarr~=2.16.1
+
```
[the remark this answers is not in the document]
leothomas: Awesome! I didn't know about that setting, thanks @chuckwondo !

environment.yml
leothomas: Did you mean conflicts with [ci/osx-arm64](https://github.com/Clay-foundation/model/tree/ci/osx-arm64), rather than conflicts with `main`? I didn't notice any conflicts with `main`. 

I merged the changes from `ci/osx-arm64` and, while I'm able to install from `conda-lock.yml` (as @chuckwondo suggested) I'm not able to install from `environment.yml` due to `cuda` being unavailable in the docker container: 
```
> [claymodel 4/4] RUN micromamba create -y -n claymodel --file environment.yml &&     micromamba clean --all --yes:
135.1 error    libmamba Could not solve for environment specs
135.1     The following package could not be installed
135.1     └─ pytorch ~=2.1.0 *cuda12* is not installable because it requires
135.1        └─ __cuda, which is missing on the system.
135.3 critical libmamba Could not solve for environment specs
------
failed to solve: process "/usr/local/bin/_dockerfile_shell.sh micromamba create -y -n claymodel --file environment.yml &&     micromamba clean --all --yes" did not complete successfully: exit code: 1
```

I'm thinking that the `conda-lock.yml` and `environment.yml` files aren't synchronized? 

Let me know what the best course of action is. I can look into making a [composite docker image](https://micromamba-docker.readthedocs.io/en/latest/advanced_usage.html#adding-micromamba-to-an-existing-docker-image), based off of both `micromamba` and [nvidia/cuda](https://hub.docker.com/r/nvidia/cuda) so that we can install pytorch with cuda support in the docker container

leothomas:
Awesome! Thank you both! I've addressed some of the request changes: 
- Regenerating the `conda-lock.yml`
- Installing the micromamba environment from `conda-lock.yml` rather than `environment.yml` 
- Removing the `pyarrow` dependency
- Adding a `.dockerignore` which ignores everything by default and only allows the files specifically needed
- Updating the `README` with relevant info on the docker image/container and how to run it
- Updating the jupyter lab flags to avoid the unnecessary usage of `--allow-root` and address some warning/debugging logs

Do y'all think it would be valuable to try to see if I can get this to run with a cuda docker image to enable the pytorch-cuda installation?

environment.yml
leothomas: Hm that's odd - there were no merge conflict with `main` in the case. I'm might have been missing something. Do the environment.yml and conda-lock.yml files look correct in their current state?

conda-lock.yml
chuckwondo, line 1: This lock file is incorrect. Your version is a "rendered" lock file usable _only_ on linux-64 platforms. It is not a _unified_ lock file.

After adding `osx-arm64` to the list of `platforms` in `environment.yml` (or for _any_ change to that file), you must regenerate `conda-lock.yml` using the following commands:

```
rm conda-lock.yml
conda lock --mamba --with-cuda=12.0
```

Note that you must first remove the current `conda-lock.yml` file because it assumes it is a unified lock file, not a "rendered" lock file (which it currently is, and shouldn't be), so it chokes on the incorrect syntax.

Please also update `README.md` as well. It has an outdated `conda-lock lock` command that includes `--platform`, which is no longer necessary now that we've added `platforms` to `environment.yml`. Note also that we don't need to explicitly specify `environment.yml` because that's the default file it looks for. Note also that `conda-lock lock` can be shortened to `conda lock`.

chuckwondo (changes requested):

.github/workflows/test.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.11"]
-        os: [ubuntu-22.04, macos-12]
+        os: [ubuntu-22.04]
```
chuckwondo, line 22: Why was macos-12 removed?

conda-lock.yml
chuckwondo: FYI, because you changed `conda-lock.yml` to be a "rendered" lock file, it is no longer a YAML file, which is why this failed pre-commit: https://results.pre-commit.ci/run/github/698210830/1709151304.3y_Z-09AQ7C3_Dgvkyf60g

If you follow my suggestion above, you will generate a new YAML file, which will fix this pre-commit failure as well.

docs/partial-inputs-flood-tutorial.ipynb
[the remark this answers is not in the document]
leothomas: The notebooks are outside the scope of what I've touched on in this PR. I'll log an issue and resolve this thread

conda-lock.yml
leothomas: @chuckwondo sorry for the slow follow up. 

I've regenerated the `conda-lock.yml` file, and removed the platform specification from the `README.md`. 

I've left the usage of the `conda-lock` since the docker image is based off of `micromamba` and doesn't have `conda` commands available (but does have `conda-lock`)

.github/workflows/test.yml
leothomas: Not sure - this was in @weiji14 's modifications in the OS-X branch

weiji14: Maybe a bad merge conflict? I didn't remove `macos-12` though, actually added `macos-14`.

docs/partial-inputs-flood-tutorial.ipynb
```diff
@@ -251,7 +251,7 @@
    "outputs": [],
    "source": [
     "outdir = Path(\"data/minicubes\")\n",
-    "outdir.mkdir(exist_ok=True, parents=True)\n",
+    "assert outdir.exists()\n",
```
weiji14, line 254: Revert changes here
```suggestion
    "outdir.mkdir(exist_ok=True, parents=True)\n",
```

docs/partial-inputs-flood-tutorial.ipynb
```diff
@@ -303,8 +303,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "DATA_DIR = \"data/minicubes\"\n",
-    "CKPT_PATH = \"https://huggingface.co/made-with-clay/Clay/resolve/main/Clay_v0.1_epoch-24_val-loss-0.46.ckpt\"\n",
+    "DATA_DIR = \"/home/ubuntu/data/minicubes\"\n",
+    "CKPT_PATH = \"https://huggingface.co/made-with-clay/Clay1/resolve/main/Clay_v0.1_epoch-24_val-loss-0.46.ckpt\"\n",
```
weiji14, line 307:
```suggestion
    "DATA_DIR = \"data/minicubes\"\n",
    "CKPT_PATH = \"https://huggingface.co/made-with-clay/Clay/resolve/main/Clay_v0.1_epoch-24_val-loss-0.46.ckpt\"\n",
```

docs/partial-inputs.ipynb
```diff
@@ -204,7 +204,7 @@
    "outputs": [],
    "source": [
     "outdir = Path(\"data/minicubes\")\n",
-    "outdir.mkdir(exist_ok=True, parents=True)\n",
+    "assert outdir.exists()\n",
```
weiji14, line 207:
```suggestion
    "outdir.mkdir(exist_ok=True, parents=True)\n",
```

docs/partial-inputs.ipynb
```diff
@@ -265,7 +265,7 @@
    ],
    "source": [
     "DATA_DIR = \"data/minicubes\"\n",
-    "CKPT_PATH = \"https://huggingface.co/made-with-clay/Clay/resolve/main/Clay_v0.1_epoch-24_val-loss-0.46.ckpt\"\n",
+    "CKPT_PATH = \"https://huggingface.co/made-with-clay/Clay1/resolve/main/Clay_v0.1_epoch-24_val-loss-0.46.ckpt\"\n",
```
weiji14, line 268:
```suggestion
    "CKPT_PATH = \"https://huggingface.co/made-with-clay/Clay/resolve/main/Clay_v0.1_epoch-24_val-loss-0.46.ckpt\"\n",
```

environment.yml
```diff
@@ -30,3 +30,4 @@ dependencies:
 platforms:
   - linux-64
   - osx-64
+  - osx-arm64
```
weiji14, line 33: Revert this line for now, since it will be added in #164
```suggestion
```

src/callbacks_wandb.py
```diff
@@ -235,4 +235,3 @@ def on_validation_end(
                 axs[1, i].axis("off")
 
             self.logger.experiment.log({"Images": wandb.Image(fig)})
-            plt.close(fig)
```
weiji14, line 238: Can you undo the line removal here manually? Can't make a suggested change on a deleted line

weiji14 (commented):
Some suggested changes to reduce diff from merge conflict handling.

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+WORKDIR /model
+
+# Address warning about python debugger missing breakpoints
+# due to frozen modules being used
+ENV PYDEVD_DISABLE_FILE_VALIDATION=1
+
+COPY --chown=$MAMBA_USER:$MAMBA_USER . .
+
+RUN micromamba create -y -n claymodel --file conda-lock.yml && \
```
weiji14, line 11: Hmm, I'm getting this error when running `docker build . -t clay --platform linux/amd64` locally:

```
0.384 Transaction starting
79.51 critical libmamba Failed to create dir 'info'
79.51 error    libmamba Error opening for reading "/opt/conda/pkgs/cudnn-8.9.7.29-h092f7fd_3/info/index.json": No such file or directory
79.51 error    libmamba Error when extracting package: [json.exception.parse_error.101] parse error at line 1, column 1: syntax error while parsing value - unexpected end of input; expected '[', '{', or a literal
79.51 cudnn-8.9.7.29-h092f7fd_3.conda extraction failed
79.73 critical libmamba Found incorrect download: cudnn. Aborting
```

Not sure if it's because something changed when I re-locked the conda-lock.yml file manually when doing the merge from main at https://github.com/Clay-foundation/model/pull/166/commits/5be9fa2fdafdd2e54692687edd1c1b50db2c12fe#diff-63113c19c5d310b8e350b302f279e2297ba6faa9ac9c99c1c82e83e508447865

leothomas: Oh jeez - I'm getting a segmentation fault: 

```bash
2.612 Transaction starting
1441.7 qemu: uncaught target signal 11 (Segmentation fault) - core dumped
1441.7 bash: line 1:    10 Segmentation fault      micromamba create -y -n claymodel --file conda-lock.yml
------
Dockerfile:11
--------------------
  10 |
  11 | >>> RUN micromamba create -y -n claymodel --file conda-lock.yml && \
  12 | >>>     micromamba clean --all --yes
  13 |
--------------------
ERROR: failed to solve: process "/usr/local/bin/_dockerfile_shell.sh micromamba create -y -n claymodel --file conda-lock.yml &&     micromamba clean --all --yes" did not complete successfully: exit code: 139
```

Wondering if it's either related to a lack of available memory or installing CUDA in docker. 

I recall that when we installed the model libraries locally we had to install pytorch without CUDA - could the CUDA installation have made its way into the conda-lock file and causing issues?

yellowcap:
Got this to work locally. Docker-compose and Jupyter notebook working fine with the image built from the Dockerfile. 🌮 

@leothomas have a look and let me know if you think the small changes make sense to you. Happy to merge this in after some review. Not sure how to use the functionality in things like binder but would be nice to use the dockerized version for those if that unbreaks the deployments!

yellowcap (approved):
🐳

yellowcap:
@chuckwondo re-requested your review to unblock this. I think with v1 it's now working ok. If you have some time try it out 🐋

yeelauren:
Hey just wanted to flag -   I noticed some of the paths for the notebook are broken in docker as well:
image

I think it should be two directories up or explicit instructions on changing your workspace directory paths.

Issue #156, opened 2024-02-18
Upper image does not train?


brunosan:
I took a 1k chips and trained it for 2k epochs. 

Based on the images saved from the [wandb hook](https://github.com/Clay-foundation/model/blob/3f4210ff3160240e73c4d3541962032afef957db/src/callbacks_wandb.py#L100), it seems clear that the upper part of the images are not really learning, or not at similar rates. As if its weights are not updated.

This is the animated video (made with [ffmpeg](https://gist.github.com/brunosan/8b4179079ec97d98861e85fd57a0a38d)).

https://github.com/Clay-foundation/model/assets/434029/ca432461-2d5f-4e46-a624-040dd1c5a02b

I'm not clear how this could happen, neither if actually a difference in the learning or on the unpatchify code. for visualization

brunosan:
I continue investigating. This pattern of not learning on the upper half seems to apply across all bands.


https://github.com/Clay-foundation/model/assets/434029/d0d4684e-adfd-42d2-9591-fd18c0ad8d53

brunosan:
I made a last test with another image set. Same pattern:

Top row is input channels, middle is output and bottom is normalized difference.


https://github.com/Clay-foundation/model/assets/434029/fb58836c-f7cd-4e2f-8ce2-4be5898f2b7c

Trying to understand this:
* Masking is 75% so I do not think that the upper part is the masked out patches that somehow also got sequential instead of random.
* The DEM (last column on the right) does converge with the top remaining random.
* For all other bands, the lower image does seem to converge, but not to the input image (are the patches scrambled).

brunosan:
If helpful, the code to create the images above is [here](https://github.com/Clay-foundation/model/blob/Image_logs/src/callbacks_wandb.py)

It selected the image on the first 3 batches with maximum variance (as a way to pick an image that has features, instead of just flat), then pulls the RGB as RGB, and the rest of the groups using average:

```python
class LogIntermediatePredictions(L.Callback):
    """Visualize the model results at the end of every epoch."""

    def __init__(self):
        """
        Instantiates with wandb-logger.
        """
        super().__init__()
        self.selected_image = None

    def on_validation_end(
        self,
        trainer: L.Trainer,
        pl_module: L.LightningModule,
    ) -> None:
        """
        Called when the validation loop ends.
        At the end of each epoch, takes the first batch from validation dataset
        & logs the model predictions to wandb-logger for humans to interpret
        how model evolves over time.
        """
[...]

PR #170, opened 2024-03-02
More detailed image plots

brunosan:
This changes create a more detailed Image log, with the actual RGB, <IR>, <SAR> and DEM.

It also chooses the most likely item of the first 3 batches to have structure (max stdev), and plots all bands, the predictions and the differences, like this

image

### DIFF
--- a/src/callbacks_wandb.py
+++ b/src/callbacks_wandb.py
@@ -30,6 +30,9 @@ import skimage
 import torch
 from einops import rearrange
 
+from itertools import islice
+
+
 try:
     import wandb
 except ImportError:
@@ -174,6 +177,7 @@ class LogIntermediatePredictions(L.Callback):
         Instantiates with wandb-logger.
         """
         super().__init__()
+        self.selected_image = None
 
     def on_validation_end(
         self,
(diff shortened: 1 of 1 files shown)


alkalait:
Hello! Where can I find a how-to on reading these plots?

brunosan:
Hi @alkalait ! I don't actually think this will get merged, but here's a description of what you see.

The code picks the item on the first 3 batches with most stdev as a proxy for features within.

Each Clay has many spectral bands, so the plot tries to see what's being learned by displaying the RGB as a color image, then the average of all red edge, Infrareds, SAR and the elevation model (all this on the top row).

the middle row has the output as the model learns to reconstruct the image. Random at first, and ideally ends up being the same as the input.

The bottom row is just the normalized difference.

In a ideal world, the middle row goes from random to the same as the first row, as the bottom row becomes grey (no difference). Hope it helps.

alkalait:
That helps, thanks @brunosan. Based on what you said, in this particular instance the model is not doing well at reconstructing the patches. Am I reading it correctly?

brunosan:
Indeed. I did not leave it long, but yes, the reconstruction is not even close. 
You basically need to leave it longer on training, but are also tracking a potential issue on #156.

yellowcap:
We plan to implement RGB plots for the v1 output, we are not planning to update v0 ouptut at this point anymore.

Issue #168, opened 2024-02-29
Bbox, date, and cloud cover for patch level embeddings


yellowcap:
To further enable patch level embeddings, two tasks:

1. Add functionality so that the `predict.py` routine outputs patch level embedding bounding box.
2. Track cloud coverage in each patch for quality filtering when using the embeddings
3. Include date stamp in the embedding database as well

This ticket is to track your request for patch level embeddings for bboxes  @danhammer

lauracchen:
Hi @yellowcap do you have a sense of when this might be possible by? We're hoping to be able to fully work with patch-level embeddings and further explore their usefulness as soon as possible to inform next steps for our app development!

lauracchen:
That's great to hear, thank you!

lauracchen:
hi @yellowcap any update here or any roadblocks? Maybe @brunosan can help out if there's any blockers.

brunosan:
not sure if already in scope, but also the option to not average the band groups.

yellowcap:
We made good progress, I will review #172 today and hopefully we can merge that soon.

lillythomas:
Yes @lauracchen I'm incorporating Daniel and Wei Ji's comments today, so #172 should be merged by Monday morning.

lillythomas:
@lauracchen patch level bounding box, date and cloud cover information have been addressed in #172 and #184 respectively, both of which are merged into main now. The notebooks to demonstrate are `docs/tutorial_digital_earth_pacific_patch_level.ipynb` and `docs/patch_level_cloud_cover.ipynb`.

They use the Digital Earth Pacific data (#140) to focus the examples.

lauracchen:
thank you so much!! @lillythomas

yellowcap:
Closing this as done, let's reopen or make a new ticket for additional questions.

PR #172, opened 2024-03-06
Obtain patch level metadata (e.g. geospatial bounds and cloud cover), save and demo DEP use case (sim search)

lillythomas:
This PR will work towards a demonstration of how to obtain patch level embeddings and write them to GeoParquet files to run similarity search with. 

Main tasks that need to be done:

- [x] Given the geospatial bounds of a 512x512 parent chip, obtain the child 32x32 geospatial bounding boxes (patches)
- [x] Map the child bounding boxes to patch level embeddings (making sure `embeddings_level = "patch"`)
- [x] Create geodataframes for the patch embeddings with their mapped bbox and write to GeoParquet
- [x] Load in a lanceDB
- [x] Use the ground truth points from #140 https://github.com/Clay-foundation/model/discussions/140#discussioncomment-8437613  [mineral extraction training data](https://github.com/digitalearthpacific/mineral-resource-detection/blob/main/training_data/draft_inputs/MRD_dissagregated_25.geojson) as reference for similarity search against all patch embeddings
- [x] compute and save patch level cloud cover percentages (+ add to db table for sim search)

Reference tickets: #168 #140


lillythomas:
The notebook `docs/tutorial_digital_earth_pacific_patch_level.ipynb` walks through an example of:
- generating patch level embeddings for an area where known mining extraction events occur
- saving the patch level embeddings to independent GeoParquet files
- executing similarity search based on a ground truth point's overlapping patch

@weiji14 @yellowcap ready for when you have time to review.

weiji14:
Thanks @lillythomas! I haven't looked too closely yet, but would it be possible to show where the similarity search results are located? Maybe something like showing the bounding boxes of all the patches on a map, and also overlay where the original quarry points are.

lillythomas:
> Thanks @lillythomas! I haven't looked too closely yet, but would it be possible to show where the similarity search results are located? Maybe something like showing the bounding boxes of all the patches on a map, and also overlay where the original quarry points are.

Yes! Great idea. Working on this tomorrow.

yellowcap (commented):
Thanks Lilly, this is great to see! I did not yet have time get to the sim search part, but will get back to it on Monday. Left some comments for now.

Could you add some more context on what the purpose is? Do you think we going to apply this to a region for searching certain events / features?

docs/tutorial_digital_earth_pacific_patch_level.ipynb
```diff
@@ -0,0 +1,1099 @@
...
...sjeo3AOg0AnpNw6VmlLlcAeCSUz6fs5M4uJfbIS9ro5asObXyMrpA4hy5GsL80fZqQm0iAOGtRlhZHq3s0EMoA3QQt1eAXv7Hw08/lUhj9lWWk5ZXXi0fYK5vNW0fTEKINMbups07TuEizL8MAELIP0eHi9YvLWFCdA91IEdCNbSHWBauJSxSCNrnqJWUqwvCgH21/KIqqVZOTg4KCgowYcKEOu+9++67MBqN6N69OwDgww8/xDvvvIOVK1cqMVOnTsWAAQOwdOlSpKamYt26ddi3bx/efvvtG7YOjDHm6/wqMaxatQp9+/ZFx44d631/4cKFOHXqFPR6PTp27Ij3338fDzzwgPJ+3759sXbtWsyfPx9//etf0b59e2zcuBGdO3e+UavAGGM+z68Sw9q1axt8b+zYsRg7dqy0jdGjR2P06NFN2S3GGPtV8ctyVcYYY9cPJwbGGGMqfnUqyZfYHQ54vFTIEMdxgyBUQmiIY195CBU0lEG7AACEihA7sbrEaJLHERYHAHC75JUVWkLFGgDoCIMA6onVZWGBgdIYJ2EgtxMlZdIYADhXLr8Df+lb/yC11Xv7NmlM/pHvSG1RaIjVMZSx4zxO2j5IKEqChzpiImHqWLeH2C/C15E6iCa8TTUMWpVXLT5iYIwxpsKJgTHGmAonBsYYYyqcGBhjjKlwYmCMMabCiYExxpgKl6teJZNWA6O24fIvammYzU4o5STEAABlWlt46fPlNBqdPIZYRucilJhSygkBIFQnL0V1Scr2agUFh8pjjLSviJ6wAuVVl6QxldXyuaMB2udzqvAMqa2zhDgjuWZSLjCQVk5M2Vf1oPWrhlA+KqhzPhPKbUnziQMQhH2VuDtDr/W+T2gE/TPkIwbGGGMqnBgYY4ypcGJgjDGmwomBMcaYCicGxhhjKlyVdJVsTgcMaLjywEiongEAjVb+EQQQB4VzE8IEsYrDYJD/zaAx0XYfLWGwOqGn9StET5gm1EMbDE1HGCDP4aK1VUOovHIRBkx0EUu9tISKHer0sm5C1YuLuB0o9BpaxQ5lfLyfq2lTe9qEvNrLGEDbn90u+WfkstPW0RAoX2aI3khqy233/hlRp/UF+IiBMcbYFTgxMMYYU+HEwBhjTIUTA2OMMRVODIwxxlT8JjEcO3YM9913H2JiYhAWFob+/ftj+/btqpiioiKkpqYiKCg
```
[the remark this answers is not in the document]
lillythomas: This was introduced by making your suggested change in https://github.com/Clay-foundation/model/pull/172#discussion_r1526415973 but I have a way to handling it. Essentially pyarrow doesn't like a polygon object, so I can get a list equivalent for the table by adding the bounds method to `"box": row["box"]` e.g. `"box": row["box"].bounds`, which I am in favor of doing as it allows us to drop the lamba function as you pointed out.

src/model_clay.py
```diff
@@ -790,7 +790,7 @@ def __init__(  # noqa: PLR0913
         wd=0.05,
         b1=0.9,
         b2=0.95,
-        embeddings_level: Literal["mean", "patch", "group"] = "mean",
```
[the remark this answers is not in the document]
lillythomas: True. Will do.

lillythomas:
@weiji14 @yellowcap when you have a moment, this could use another review before merging.

docs/tutorial_digital_earth_pacific_patch_level.ipynb
```diff
@@ -0,0 +1,1311 @@
...
+    "    items,\n",
+    "    bounds=bounds,\n",
+    "    snap_bounds=False,\n",
+    "    epsg=epsg,\n",
+    "    resolution=10,\n",
+    "    dtype=\"float32\",\n",
+    "    rescale=False,\n",
+    "    fill_value=0,\n",
+    "    assets=BAND_GROUPS[\"rgb\"] + BAND_GROUPS[\"nir\"],\n",
+    "    resampling=Resampling.nearest,\n",
```
weiji14, line 369: Since we're working with optical imgaery, it is recommend to set pixel registation to center registered, rather than `stackstac`'s default topleft (gridline) registration. See https://stackstac.readthedocs.io/en/latest/api/main/stackstac.stack.html#stackstac.stack.params.xy_coords.

```suggestion
    "    resampling=Resampling.nearest,\n",
    "    xy_coords=\"center\",\n",
```

weiji14 (commented):
Could you please add this tutorial notebook to `docs/_toc.yml`, so that the notebook will render on merge? Also made some suggested changes/typos below.

One major issue I saw is on conflating Digital Earth Pacific's cloud-free GeoMAD composite with a clouded Sentinel-2 image. Matching SCL metadata from raw Sentinel-2 images to the GeoMAD composite doesn't seem correct...

docs/tutorial_digital_earth_pacific_patch_level.ipynb
```diff
@@ -0,0 +1,1311 @@
...
+    "Geometric Median [(GeoMAD)](https://docs.digitalearthafrica.org/en/latest/data_specs/GeoMAD_specs.html) composite. We will use\n",
+    "[Digital Earth Pacific's STAC API](https://stac-browser.staging.digitalearthpacific.org)\n",
+    "to obtain these datasets, and apply it on a mineral resource detection downstream task to do the following:\n",
+    "\n",
+    "1. Generate fine level (pixel of size 10m x 10m) embeddings for an area (5.12km x 5.12km) where known mining extraction is occurring\n",
+    "2. Save the fine level (patch) embeddings and execute a similarity search that leverages the ground truth points as reference\n",
+    "\n",
+    "We will also leverage the [AWS Sentinel-2 STAC catalog](https://registry.opendata.aws/sentinel-2/) to obtain fine \"patch\" level cloud cover percentages from the [Scene classification (SCL) mask](https://docs.sentinel-hub.com/api/latest/data/sentinel-2-l2a/). These percentages will be mapped and added to the GeoParquet files such that they can be added to database tables for similarity search filters and any other relevant downstream tasks.\n",
```
weiji14, line 18: Mixing the cloud-free composite from DEP'S STAC with the SCL mask from AWS's Sentinel 2 STAC seems... weird? I see that you're trying to consolidate the notebook from #184 with the one here, but there's an inherent mismatch with the data from two different STAC catalogs. It won't make sense to do a cloud-percentage pre-filter using the SCL band, when the composite image is technically cloud-free already.

docs/tutorial_digital_earth_pacific_patch_level.ipynb
```diff
@@ -0,0 +1,1311 @@
...
+    "\n",
```
weiji14, line 19: Mixing the cloud-free composite from DEP'S STAC with the SCL mask from AWS's Sentinel 2 STAC seems... weird? I see that you're trying to consolidate the notebook from #184 with the one here, but there's an inherent mismatch with the data from two different STAC catalogs. It won't make sense to do a cloud-percentage pre-filter using the SCL band, when the composite image is technically cloud-free already.

yellowcap: For this example notebook, could we use a regular Sentinel 2 scene from AWS? We could hand pick one that does not have many clouds, but maybe a few as to demonstrate the cloud percentage calculation?

In other words, what is the advantage of using the DEP STAC in this example?

Another solution could be to use a light weight cloud detection model that could compute a cloud mask based on the bands directly. There are a few, but not sure how easy they are to use.

lillythomas:
Yes @weiji14 @yellowcap I didn't intend to suggest that we would use the derived cloud percentages to filter the embedding results obtained from cloud-free DEP composite, just that the two leveraged some similar things (i.e. AOI) and the initial ask for patch-level metadata https://github.com/Clay-foundation/model/issues/168 was unified so I sought to consolidate.

I guess indeed if we want this to be a sensible consolidation, we could demonstrate this without the cloud-free composite or I can just re-separate the PRs. Whatever is swiftest and most useful. I'll address some of the other comments from Wei Ji first then comment on which of the two directions we go with as far as the above is concerned.

lillythomas:
Re-separated into different tutorials / PRs. See https://github.com/Clay-foundation/model/pull/184 for the cloud cover tutorial.

yellowcap:
I was able to run the notebook, works nicely now! Only one celll copmlained, the one with the plot of the similar patches, I get the following error.

```
ValueError                                Traceback (most recent call last)
Cell In[53], line 41
     39 idx = row["idx"]
     40 # Find the corresponding window based on the idx
---> 41 window_index = idxs_windows["idx"].index(idx)
     42 window_data = idxs_windows["window"][window_index]
     43 # print(window_data.shape)

ValueError: '20180828_4_2' is not in list
```

I really like the last plot showing the location of the search results. I would have one request there: would it be possible to highlight which one was the input patch? For the user to know what the reference was.

I thought of this because in the plot none of the patches have a ground truth point right in the center. At least the reference chips should be right on top of a quarry ideally.

lillythomas:
Thanks, glad you could reproduce @yellowcap. As for that error you're encountering, we shouldn't have any dates in 2018 in this tutorial (see `daterange: dict = ["2021-01-01T00:00:00Z", "2021-12-31T23:59:59Z"] `)  so maybe just double check `data/minicubes` is cleared before running the notebook. Other than that, I'll adjust the plot to support that request; I agree it would be useful.

lillythomas:
@yellowcap specifying the reference patch in the final plot was achieved in commit  https://github.com/Clay-foundation/model/commit/1f9bdf53f3a3811f08aeb7f205f2d1cfd2df0dc1.

> At least the reference chips should be right on top of a quarry ideally.

The reference patch overlaps directly with a ground truth point. It is the box in yellow here:
Screen Shot 2024-03-21 at 5 17 41 PM

yellowcap:
Looks like the linter is not happy about the size of the notebook.

I would propose to replace the cells with images with links that will make the notebook smaller.

I did this by uploading the images from the cells to the PR or Issue, then copy the link and place it in a markdown cell. The following is an example:
```
![Minicube visualization](https://github.com/Clay-foundation/model/assets/901647/c6e924e5-6ba1-4924-b99a-df8b90731a5f)
```

See https://github.com/Clay-foundation/model/blob/main/docs/partial-inputs.ipynb?short_path=984f94c#L181

yellowcap (approved):
This is good to go once the linters are happy! 🚀

Issue #178, opened 2024-03-12
Clay v1 Pipeline Spec


yellowcap:
This is a proposal for a streaming approach that relies on static STAC Indexes and dynamic tiling.

### tl;dr
- List imagery "scenes" using STAC items
- Compute how many chips are in each scene
- Download batches of "scenes" to a fast temporary storage
- Do dynamic tiling for feeding to the model

## Tracking scenes

The first step is creating a list of scenes from multiple sources, like Modis, Landsat, Sentinel-2, Naip, Drones. Separately, one by one. 

* For each scene create a **static** STAC item with asset links to scalable storage (S3), ideally in same region
* Each STAC item metadata should contain the following information
    * Reference date
    * Central wavelength of each band
    * Bandwidth of each band
    * Resolution in meters of each band
    * lat/lon bounding box
    * Data normalization average
    * Data normalization mean

## Track chips in each scene

Track how many chips are in each scene

* For each STAC item, we precompute how many “chips” can be created using simple window reads. That means no reprojection or spatial lookup, simply cutting the pixels into chips of 256x256 pixels
* With this count we can index all the chips and their “parent” STAC items

### The index table
We would then store all the static STAC item as json in files in a central location and index all the chips into a single table.

<table>
  <tr>
   <td>Stac Item
   </td>
   <td>Chip ID
   </td>
  </tr>
  <tr>
   <td>sentinel-2-2023-01-23-FQ23.json
   </td>
   <td>1
   </td>
  </tr>
  <tr>
   <td>sentinel-2-2023-01-23-FQ23.json
   </td>
   <td>2
   </td>
  </tr>
  <tr>
   <td>sentinel-2-2023-01-23-FQ23.json
   </td>
   <td>3
   </td>
  </tr>
  <tr>
   <td>…
   </td>
   <td>…
   </td>
  </tr>
  <tr>
   <td>naip-ma-2020-m-41071327.json
   </td>
   <td>1301
   </td>
  </tr>
  <tr>
   <td>naip-ma-2020-m-4107132.json
   </td>
   <td>1302
   </td>
  </tr>
  <tr>
   <td>naip-ma-2020-m-4107132.json
   </td>
   <td>1303
   </td>
  </tr>
  <tr>
   <td>…
   </td>
   <td>
   </td>
  </tr>
  <tr>
   <td>openaerialmap-65ec6c0deda0.json
   </td>
   <td>2521
   </td>
  </tr>
  <tr>
[...]

PR #180, opened 2024-03-14
First try for v1 pipeline

yellowcap:
This is a proof of concept for #178. This is kind of a generic approach that is solely based on STAC, so we could spin this out into a separate pipeline repository and make it standallone for it to be used also in other projects.

Its functional but a bit clunky, but it does show how this could work.

To test it locally:

Create a working directory with an `/items` subdirectory.

1. In there, run `create_stac.sh`, this should create local static STAC items
2. Then run `edit_stac.py` this will change default source paths for Landsat to S3 locations.
3. Then run `indexing.py` this should create an index json file that represents the items and the chips in them
4. Then finally the `chipping.py` script should print chips that are dynamically created from the index.

### DIFF
--- a/scripts/pipeline_v1/chip_indexer.py
+++ b/scripts/pipeline_v1/chip_indexer.py
@@ -1,0 +1,160 @@
+import warnings
+from pathlib import Path
+from typing import Tuple
+
+import numpy as np
+import pyarrow as pa
+import stackstac
+from pystac import Item
+
+warnings.filterwarnings(
+    "ignore",
+    message=(
+        "The argument 'infer_datetime_format' is deprecated and will"
+        " be removed in a future version. A strict version of it is now "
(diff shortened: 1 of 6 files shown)


yellowcap:
I have been working more on the v1 pipeline, realizing that we will need cloud and nodata tracking at the chip level. I played around with creating classes where the platform specific cloud and nodata filters can be added through class method overrides.

the classes are at

https://github.com/Clay-foundation/model/blob/clay-v1-pipeline/scripts/pipeline_v1/chip_indexer.py

script to try them out with the data attached below is 

https://github.com/Clay-foundation/model/blob/clay-v1-pipeline/scripts/pipeline_v1/run_chip_indexer.py

[test_stac_items.zip](https://github.com/Clay-foundation/model/files/14725812/test_stac_items.zip)

yellowcap:
This has migrated to 
https://github.com/Clay-foundation/stacchip

PR #198, opened 2024-03-27
Updating location to download imagery script in docs

MaxLenormand:
Noticed the bash script in `docs/run_region.md` didn't point to the correct location anymore so updating to [point to the new location of `datacube.py`](https://github.com/Clay-foundation/model/blob/main/scripts/pipeline/datacube.py)


weiji14 (approved):
Great catch, thanks!

PR #214, opened 2024-04-09
Change "Radiant Earth Foundation" to "Radiant Earth" in docs

kbgg:
We've dropped the "Foundation" from our name


weiji14 (approved):
Thank you!

PR #221, opened 2024-04-17
Template for Clay V1 model

srmsoumya:
- Add support for dynamic embeddings from DOFA paper
- Move time & latlon embedding to the encoding step
- Add cls token to learn representation per chip


srmsoumya:
Merged, as I need to implement the continual learning framework on top of this.

PR #193, opened 2024-03-26
Fixes issue with model reconstruction of the upper half of the image & saves model checkpoint in s3

srmsoumya:
This PR resolves an issue with the model reconstructing just the bottom 50% of the image during validation and stores model checkpoints in the s3 store.
- Adds a shuffle argument to ClayModule that is set to False by default
- Logs model checkpoints to aws s3 bucket

Fixes #156 #138

PR #225, opened 2024-04-21
Update instructions to re-lock conda-lock.yml file

weiji14:
Remove the `--platform linux-64` flag since unified lockfile is for linux-64, osx-64 and osx-arm64 as of #164. Also re-locking the conda-lock.yml file after 2a9ef9d6d73bd0c20b0ce04d32b6f1cfac8bd5cf/#193.

Patches #164, resolves https://github.com/Clay-foundation/model/pull/193#issuecomment-2066410721.

### DIFF
--- a/conda-lock.yml
+++ b/conda-lock.yml
@@ -13,9 +13,9 @@
 version: 1
 metadata:
   content_hash:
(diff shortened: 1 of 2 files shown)


weiji14:
Merging directly to fix broken tests on `main` branch.

Issue #94, opened 2023-12-19
Switch from per-band mean/std normalization of Sentinel-2 values to surface reflectance conversion instead


weiji14:
Using mean and standard deviation normalization is a common procedure in standard Computer Vision, and can be applied e.g. by using `torchvision`'s [Normalize](https://pytorch.org/vision/main/generated/torchvision.transforms.Normalize.html) function. But this normalization can lead to incorrect band ratios when applied to optical remote sensing images.

E.g. let's take the formula for Normalized Difference Vegetation Index (NDVI):

$$
\text{NDVI} = \frac{\text{NIR} - \text{Red}}{\text{NIR} + \text{Red}}
$$

If say, we have a un-normalized Sentinel-2 pixel with Band 8 (NIR): 3327, and Band 4 (Red): 426, then the NDVI value would be:

$$
\text{NDVI} = \frac{3327-426}{3327+426} = 0.77
$$

However, if we apply a per-band mean/std normalization scheme, the value becomes:

```math
\text{NIR} = \frac{3327-2238}{1414} = 0.77
```
```math
\text{Red} = \frac{426-583}{981} = -0.16
```
```math
\text{NDVI} = \frac{0.77-(-0.16)}{0.77+(-0.16)} = 1.52
```

Clearly this is wrong, since we've removed the NDVI signal! A model trained on these mean/std normalized pixel values would have a harder time capturing the semantics of band indices such as NDVI.

One possible solution, is that instead of applying a per-band normalization, we can convert the Sentinel-2 Digital Number (DN) values to surface reflectance by dividing with the dynamic range of the band to a value between 0-1. Sentinel-2's MSI sensor is 12-bit, but the data is stored as 16-bit. Usually people use 10000, but this doesn't work for very bright white areas, so I'll use $2^{14} = 16384$ below:

```math
\text{NIR} = \frac{3327}{16384} = 0.20
```
```math
\text{Red} = \frac{426}{16384} = 0.026
```
```math
\text{NDVI} = \frac{0.20-0.026}{0.20+0.026} = 0.77
```
which matches with the actual NDVI value.

Notes:
- Calculating the actual surface reflectance is a little more complicated than the above for the Sentinel-2 L2A product since we also need to apply an offset (see https://sentinel.esa.int/web/sentinel/technical-guides/sentinel-2-msi/level-2a-algorithms-products), but the main point is - **don't use mean/std normalization**!
- The SCL band should be divided by a different value compared to the other optical bands (B02-B12).

[...]

PR #240, opened 2024-04-30
Improved metadata

yellowcap:
This needs another sanity check, and the actual band stats are in computation right now. Should update tomorrow morning with the stats from the calculations on batch.


configs/metadata.yaml
```diff
@@ -133,5 +167,20 @@ linz:
       red: 25
     wavelength:
       blue: 0.48
```
weiji14, line 169: Camera used for LINZ aerial surveys differs by council, but for https://data.linz.govt.nz/layer/115053-christchurch-0075m-urban-aerial-photos-2023/metadata/dc which used a Leica ADS100 line scanning camera, I got these specs off https://leica-geosystems.com/-/media/files/leicageosystems/products/datasheets/leica_ads100_ds.pdf?sc_lang=en-us&hash=8A2985763B14A56C278FA85CB5FF88C8.

image

```suggestion
      blue: 0.465
```

configs/metadata.yaml
```diff
@@ -133,5 +167,20 @@ linz:
...
-      green: 0.53
-      red: 0.66
+      green: 0.56
+      red: 0.65
```
weiji14, line 171:
```suggestion
      green: 0.555
      red: 0.635
```

configs/metadata.yaml
```diff
@@ -39,51 +39,86 @@ s2:
       swir16: 450
       swir22: 480
     wavelength:
-      blue: 0.4905
-      green: 0.5605
+      blue: 0.49
```
weiji14, line 42: Only one nitpick change for Sentinel-2, referencing https://sentinels.copernicus.eu/web/sentinel/user-guides/sentinel-2-msi/resolutions/spectral. Other bands look ok.

image

```suggestion
      blue: 0.493
```

configs/metadata.yaml
```diff
@@ -39,51 +39,86 @@ s2:
...
-l8:
+landsat-c2l1:
   band_order:
-    - blue
+    - red
     - green
+    - blue
+    - nir08
+    - swir16
+    - swir22
```
weiji14, line 59: Why is the near-infrared band for Landsat called nir08? Band 8 on Landsat 8/9 is the panchromatic band. Is this to match with Sentinel-2? Maybe we could follow the HLS naming scheme at https://lpdaac.usgs.gov/data/get-started-data/collection-overview/missions/harmonized-landsat-sentinel-2-hls-overview/#hls-spectral-bands if we want to keep the same band name. I.e. use NIR Broad, NIR Narrow, SWIR 1, SWIR 2.

weiji14 (commented):
Thanks @yellowcap, I'm just cross-checking the wavelengths for now which look mostly ok. I do have a comment/suggestion about how we might want to consistently name the Near-infrared band(s) across Sentinel-2 and Landsat (and potentially NAIP).

configs/metadata.yaml
```diff
@@ -133,5 +167,20 @@ linz:
       red: 25
     wavelength:
       blue: 0.48
-      green: 0.53
-      red: 0.66
+      green: 0.56
+      red: 0.65
+sentinel-1-rtc:
```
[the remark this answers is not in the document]
yellowcap: RTC

configs/metadata.yaml
yellowcap: The naming comes as-is from the underlying STAC catalog. We have this baked into the code in 2-3 places now. So let's not change it right now and make this more consistent in a future iteration.

Example:

https://landsatlook.usgs.gov/stac-server/collections/landsat-c2l1/items/LC09_L1GT_173239_20240430_20240430_02_T2

configs/metadata.yaml
```diff
@@ -133,5 +167,20 @@ linz:
...
+  band_order:
+    - vv
+    - vh
+  gsd: 10
```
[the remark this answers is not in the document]
yellowcap: Very good point. Maybe this could be renamed to pixel spacing. I think this is what the model need to know the most here, what size a single pixel represents. The GSD would be good additional information, but it would have to be specified per band for multiple systems. For instance, for Sentinel-2 we resample everything to a pixel spacing of 10m, although some bands have 20m GSD. So lets keep it as is for now and update naming later?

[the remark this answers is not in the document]
weiji14: Sure, we can keep the naming for now (otherwise the model code needs to be updated too).

configs/metadata.yaml
weiji14: Ah I see, then let's keep this as is for now then.

weiji14 (approved):
Pre-approving to move this along (you'll need to fix the merge conflicts first though). Still think we need to move away from mean/std normalization (https://github.com/Clay-foundation/model/issues/94), but could do it for model v2 I guess :slightly_smiling_face:

yellowcap:
This has been merged in https://github.com/Clay-foundation/model/pull/253

PR #249, opened 2024-05-14
Documented v1 sampling strategy

yellowcap:
Extended docs on sampling to describe v1 strategy.

### DIFF
--- a/docs/data_sampling.md
+++ b/docs/data_sampling.md
@@ -1,6 +1,43 @@
-# Data sampling strategy
+# Training Data
 
-To create a balanced dataset for model training, we used a sampling strategy
+This section describes how we created the training dataset for the clay model.
+
+## Data sources selection
+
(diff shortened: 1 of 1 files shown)


docs/data_sampling.md
```diff
@@ -26,8 +63,87 @@ After selecting MGRS tiles for each of these criteria, we removed duplicates.
...
+Moss and lichen | 50 | 500
+Cropland | 800 | 3600
+Tree cover | 150 | 750
+Shrubland | 100 | 500
+Grassland | 200 | 500
+Bare / sparse vegetation | 50 | 500
+Snow and Ice | 25 | 500
+Permanent water bodies | 50 | 1000
+
+This resulted in a sample of 2728 MGRS tiles total in our sample. The resulting sample file can be downloaded from the following link
```
srmsoumya, line 82: Should we add some details on how we filtered these tiles based on cloud cover percentage?

srmsoumya (approved):
Overall looks great to me, I have made a few suggestions to add if you think they are useful.

docs/data_sampling.md
yellowcap: We actually did not use that at all for selecting the 2728 MGRS locations. The cloud filtering is done for each system when selecting specific scenes at each MGRS tile location. That is documented below.

srmsoumya: ak, okay!

PR #250, opened 2024-05-21
Update specification.md

brunosan:
Closes https://github.com/Clay-foundation/office/issues/85

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -145,7 +145,7 @@ inputs it will be necessary to subset these as shown in the partial input tutori
 * Training Time:
     * `25` epochs, each taking ~`15h` to train.
 * Carbon Emissions:
-    * According to the "Customer Carbon Emission Tool", there were no emissions. Following the [documentation](https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/ccft-estimation.html), we believe this is due to the usage of renewable energy sources.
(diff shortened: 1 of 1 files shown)


docs/specification.md
```diff
@@ -145,7 +145,7 @@ inputs it will be necessary to subset these as shown in the partial input tutori
 * Training Time:
     * `25` epochs, each taking ~`15h` to train.
 * Carbon Emissions:
-    * *Report not yet available from provider, expected March'24*
+    * According to the "Customer Carbon Emission Tool", there were no emissions. Following the [documentation](https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/ccft-estimation.html), we believe this is due to the usage of renewable energy sources.
```
yellowcap, line 148:
```suggestion
    * According to the "Customer Carbon Emission Tool", there were no Scope 1 or Scope 2 carbon emissions. Following the [documentation](https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/ccft-estimation.html), we believe this is due to the usage of renewable energy sources. We are aware that Scope 3 emissions might be significant for data centers and that these are not included in the estimate. 
```

yellowcap (changes requested):
Suggested changes to be explicit about the emissions accounted for in the AWS reports. Zero emissions is too good to be true, so let's be more precise here. Scope 3 emissions (from producing the computers in the data centers)  might be significant and are not included.

yellowcap (approved):
🐈

PR #252, opened 2024-05-24
Updates the mamba install instructions link.

dbonomo:
This is a fix to the link to the Mamba install documents, which appear to have been moved. I took a look at the wayback machine for the original URL last year when it was still active, and the link I proposed appears to go to the spiritual successor of the content of the original URL.


yellowcap (approved):
Great thanks @dbonomo !

PR #253, opened 2024-05-24
Datamodule & Model for Clay v1

srmsoumya:
- Add support batch loading of data from different sensors
- Dynamic Embeddings to patchify chips from different sensors
- Add teacher model to better align the model embeddings
- Add support to feed GSD & time/latlon encodings


yellowcap (approved):
Here goes the model! 🐠

PR #256, opened 2024-05-27
Remove clay-v0-tutorials & add visualize embeddings for v1

srmsoumya:
- Remove deprecated tutorial notebooks for clay-v0
- Add a tutorial notebook to show how embedding space looks like for Clay Encoder (v1)
- Show examples for 2 sensors, i.e NAIP & Sentinel-2


lauracchen:
hey @srmsoumya is there any reason to keep but "archive" the older tutorial notebooks? I doubt there's a reason someone wants to use v0 over v1 at this point, but maybe in the future we should think about keeping deprecated notebooks so people can still make use of the older model versions if they want to?

yellowcap (approved):
Looks good 🐧

yellowcap:
> hey @srmsoumya is there any reason to keep but "archive" the older tutorial notebooks?

@lauracchen the notebooks stay in the repo, so that is definitively a way to archive them. We were also thinking of re-organising the content once we have merged the main pieces and we see what structure would make sense. Let's consider linking to the old notebooks at that point.

yellowcap:
Merging to have the updates notebooks out asap. Let's figure out the linters afterwards.

PR #261, opened 2024-06-04
Shorten comment line length

yellowcap:
This should fix the linter errors.


yellowcap:
Uploading images here that can be substituted as links in notebooks

embeddings-naip-rgb
embeddings-naip-abstract
embeddings-naip-rgb-abstract-combo
embeddings-sentinel-2-rgb
embeddings-sentinel-2-abstract
embeddings-sentinel-2-rgb-abstract-combo
reconstruction

srmsoumya (approved):
As long as we keep the clayinters happy! Thanks @yellowcap

PR #262, opened 2024-06-04
Refactor docs by moving v0 docs into separate section

yellowcap:
This moves all v0 doc files into a section that is collapsed by default.

It also slightly simplifies the docs structure.

For now some sections remain empty, but I think this is preferrable right now over having the old docs there. This sets up the structure, we can add the updated texts in the coming days.


srmsoumya (approved):
WIP but approving to keep the docs updated for Clay v1.

PR #264, opened 2024-06-05
Unified v1 and v0 environments

yellowcap:
The osx conda-lock builds are failing. Skipping them for now, let's get the updated version out the door and reactivate osx lock files in a second step.


srmsoumya (approved):
Looks good - Might have to update the docs for windows/mac.

PR #268, opened 2024-06-06
Fix duplicate transform bug

srmsoumya:


finetune/segment/chesapeake_datamodule.py
```diff
@@ -46,7 +46,9 @@ def __init__(self, chip_dir, label_dir, metadata, platform):
         )
 
         # Load chip and label file names
-        self.chips = [chip_path.name for chip_path in self.chip_dir.glob("*.npy")]
+        self.chips = [chip_path.name for chip_path in self.chip_dir.glob("*.npy")][
+            :2000
```
yellowcap, line 50: Maybe give this magic number limit a name, or even make it an init variable, so that the user understand what it is.

yellowcap (approved):
Small suggestion, but ok to go 🐑

finetune/segment/chesapeake_datamodule.py
srmsoumya: Actually, we don't have to set it for training on the full set, I will remove this & merge.

PR #281, opened 2024-06-19
Remove model_clay.py

srmsoumya:
`src/model_clay.py` has deprecated code for model v0, might confuse new users coming to clay.


yellowcap (approved):
I had the same thought yesterday. Though this will break the v0 tutorials, shall we add a comment in the docs somewhere that the old docs only work with the 0.1 tag?

srmsoumya:
@yellowcap yes, we can add an overview section for `v0` docs, that mentions the tag release users should use to pull the model. Also, recommending to use v1 over v0.

PR #286, opened 2024-06-29
Point latest release to v1.0

fnands:
Hi,

Just a small error in the docs: The latest release being pointed to is wrong. 

This PR changes it so that the docs point to release v1.0 of the model instead of v.0.0.1

Let me know if you want me to add a line in the changelog, but it's such a trivial change that it might be too small to go in.


yellowcap (approved):
Great catch! Thanks @fnands

PR #292, opened 2024-07-05
Add example for simple classification on embeddings

yellowcap:
This notebook should be fully reproducible. It download the required data from huggingface.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1207727410748364


srmsoumya (approved):
Looks neat & well done!

PR #295, opened 2024-07-09
Fix relative paths

yellowcap:
Small fixes of paths in notebook


srmsoumya (approved):
Note: May be we can push config.yaml & metadata.yaml to HF model repository.

yellowcap:
> Note: May be we can push config.yaml & metadata.yaml to HF model repository.

Good idea, tracking in #296

PR #298, opened 2024-07-10
Add no sign request flag to segment scripts and readme

yellowcap:
Thanks for your review and input @cstner


srmsoumya (approved):
Thanks for finding this.

PR #303, opened 2024-07-18
Add scripts to export the encoder of Clay

srmsoumya:


src/utils.py
```diff
@@ -11,6 +11,7 @@ def posemb_sincos_2d(h, w, dim, temperature: int = 10000, dtype=torch.float32):
     assert (dim % 4) == 0, "feature dimension must be multiple of 4 for sincos emb"
     omega = torch.arange(dim // 4) / (dim // 4 - 1)
     omega = 1.0 / (temperature**omega)
+    omega = omega.to(y.device)
```
weiji14, line 14: Can omega just be instantiated with `y.device` directly?

```suggestion
    omega = torch.arange(dim // 4, device=y.device) / (dim // 4 - 1)
    omega = 1.0 / (temperature**omega)
```

lauracchen:
@srmsoumya We are still missing the onnx GPU export right? I'm only able to run onnx for some reason, so it'd be great to have access to the GPU version too (and I'm guessing maybe others might have the same experience)

PR #319, opened 2024-08-13
Added Planetscope SR mapping

jonasViehweger:
This adds mappings for the 8-band Planetscope Surface Reflectance product. 

The statistics were calculated using the openly (CC-BY-NC) available [Planet Sandox data](https://collections.sentinel-hub.com/planet-sandbox-data/). 

Due to the available data in Planet Sandox, water and arid regions are underrepresented in the sample. This might skew the results slightly (i.e. lower std than if it was a truly random global sample).

### DIFF
--- a/configs/metadata.yaml
+++ b/configs/metadata.yaml
@@ -49,6 +49,49 @@ sentinel-2-l2a:
       nir08: 0.865
       swir16: 1.61
       swir22: 2.19
+planetscope-sr:
+  band_order:
+    - coastal_blue
+    - blue
+    - green_i
+    - green
+    - yellow
+    - red
+    - rededge
+    - nir
+  rgb_indices:
+    - 5
+    - 3
+    - 1
+  gsd: 5
+  bands:
+    mean:
+      coastal_blue: 1720.
+      blue: 1715.
+      green_i: 1913.
+      green: 2088.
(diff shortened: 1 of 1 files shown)


yellowcap (commented):
Awesome thanks @jonasViehweger . This will be super userful for many users I am sure. Would you be able to share a script on how this was calculated for future reference?

jonasViehweger:
The notebook to calculate mean and std is here: https://github.com/planetlabs/notebooks/blob/test-clay/jupyter-notebooks/workflows/planet_sandbox_data/clay-embeddings/utils/ps-mean-std.ipynb

It uses the polygons from the sandobx data which are available here: https://collections.sentinel-hub.com/planetscope/sandbox-data.html

The notebook uses Planet Insights Platform (PIP) to derive the statistics. Let me know if you think the notebook should be added to this PR and if yes, where it should be put.

I also modified your wall to wall notebook to stream the planet sandox data from PIP to generate embeddings: https://github.com/planetlabs/notebooks/blob/test-clay/jupyter-notebooks/workflows/planet_sandbox_data/clay-embeddings/clay-embeddings.ipynb

yellowcap (approved):
Very happy to see another dataset added, this should makeing Planet data much simpler for many users! 🗺️

PR #321, opened 2024-08-24
Correct LINZ resolution in metadata

wboykinm:
A quick bump to the instrument config for LINZ, which has a 5cm resolution, not 50cm.

### DIFF
--- a/configs/metadata.yaml
+++ b/configs/metadata.yaml
@@ -198,7 +198,7 @@ linz:
     - 0
     - 1
     - 2
-  gsd: 0.5
+  gsd: 0.05
   bands:
     mean:
       red: 89.96


configs/metadata.yaml
```diff
@@ -198,7 +198,7 @@ linz:
     - 0
     - 1
     - 2
-  gsd: 0.5
+  gsd: 0.05
```
weiji14, line 201: The spatial resolution of LINZ imagery varies depending on whether it is rural or urban imagery, and also the year it was collected. E.g. https://data.linz.govt.nz/layer/51769-auckland-05m-rural-aerial-photos-2010-2012/ is actually 0.5m, and the most precise urban imagery can be 0.05m (e.g. [Kaikoura 2023](https://data.linz.govt.nz/layer/119471-kaikoura-005m-urban-aerial-photos-2023/)) as you mention, but is typically more like 0.075m. I'm guessing we went with 0.5m because even the higher resolution products were resampled to 0.5m.

wboykinm: Ahhhhh, totally makes sense. Might be worth focusing in on regions by resolution, but definitely best to upscale across the country.

Thanks; sorry about the false alarm!

PR #322, opened 2024-08-24
Adjust s5cmd invocation flags

wboykinm:
For reasons I don't claim to understand, `s5cmd` requires that the `--no-sign-request` flag be positioned _before_ the `cp` command.


srmsoumya (approved):
Thank you, for bringing this to our attention.

PR #337, opened 2024-12-04
correct link to v1.5 ckpt

rbavery:
there's a link in the docs that points to an old checkpoint, when clicked it 404s


brunosan (approved):

brunosan:
❤️

PR #350, opened 2025-05-08
All for Satellogic EarthView

brunosan:
Adding input class for Satellogic EarthView.
Based on [Satellogic Documentation](https://developers.satellogic.com/imagery-products/ortho.html) and the [data release](https://satellogic-earthview.s3.us-west-2.amazonaws.com/index.html) we believe this should be called `satellogic-MSI-L1D`. There is some difference in wavelengths between Mark versions but doesn't seem too big.


jmarintur:
Hey @brunosan ! Long time no see. Yes, let's call it MSI L1D. The only thing to take into account, since v1.0 band order changed from RGBN to BGRN. 

Regarding differences in wavelengths between generations, yes you're right, there are not too big.

brunosan:
>since v1.0 band order changed from RGBN to BGRN.

Do you mean that EarthView is RGBN, but you've changed since? Might be worth then calling the set `MSI-L1D-earthview` RGBN and then create another one `MSI-L1D` with BGRN.

Please clarify the order for the EarthView set. 🙏

dingzhiyue (approved):

PR #352, opened 2025-05-19
Fix deprecated version of `actions/upload-artifact: v3`

brunosan:
Trying to fix  https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

by pinning V2


yellowcap (approved):
Lets roll!

